PR #66, opened 2018-12-21
FIX #62

### LINKED ISSUE #62: -K flag ignored
Recent builds ignore the -K flag and ask you to enter path to the keyfile.

Command line flags should override.

On entering through the prompt, the input does not properly expand ~ and concats to CWD.

aricart:
FIX issue where -K was reset in favor of expected keypath
FIX ~ paths entered in interactive where the ~ are not expanded

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -128,10 +128,15 @@ func ReadJson(fp string, v interface{}) error {
 }
 
 func Read(fp string) ([]byte, error) {
-	afp, err := filepath.Abs(fp)
+	hfp, err := homedir.Expand(fp)
+	if err != nil {
+		return nil, fmt.Errorf("error expanding path %q: %v", fp, err)
+	}
+	afp, err := filepath.Abs(hfp)
 	if err != nil {
 		return nil, fmt.Errorf("error converting abs %q: %v", fp, err)
 	}
+
 	return ioutil.ReadFile(afp)
 }
 
--- a/cmd/signerparams.go
+++ b/cmd/signerparams.go
@@ -42,23 +42,23 @@ func (p *SignerParams) Edit(ctx ActionCtx) error {
 	var err error
 
 	sctx := ctx.StoreCtx()
-	p.signerKP, err = sctx.ResolveKey(p.kind, KeyPathFlag)
-	if err != nil {
-		return err
-	}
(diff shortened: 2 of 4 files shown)


cmd/common.go
```diff
@@ -128,10 +128,15 @@ func ReadJson(fp string, v interface{}) error {
 }
 
 func Read(fp string) ([]byte, error) {
-	afp, err := filepath.Abs(fp)
+	hfp, err := homedir.Expand(fp)
```
derekcollison: This needed if we do Abs below?

cmd/signerparams.go
```diff
@@ -42,23 +42,23 @@ func (p *SignerParams) Edit(ctx ActionCtx) error {
...
-	ks := sctx.KeyStore
-	switch p.kind {
-	case nkeys.PrefixByteOperator:
-		KeyPathFlag = ks.GetOperatorKeyPath(sctx.Operator.Name)
-	case nkeys.PrefixByteAccount:
-		KeyPathFlag = ks.GetAccountKeyPath(sctx.Account.Name)
-	case nkeys.PrefixByteCluster:
-		KeyPathFlag = ks.GetClusterKeyPath(sctx.Cluster.Name)
+
+	if KeyPathFlag == "" {
```
derekcollison: If this overrides others maybe do this first and return so its clearer what we are doing?

cmd/store/keys.go
```diff
@@ -301,11 +302,15 @@ func Match(pubkey string, kp nkeys.KeyPair) bool {
 }
 
 func keyFromFile(path string) (nkeys.KeyPair, error) {
-	path, err := filepath.Abs(path)
+	hfp, err := homedir.Expand(path)
```
derekcollison: I thought Abs would expand relative but maybe not? What is homedir? Is this another dependency?

aricart: We already had the dependency - yes - abs doesn't resolve ~.

cmd/signerparams.go
aricart: It cannot some other operation could have set the default - (setting the account name via a flag, etc). So this check is delayed until the last possible moment during interactive.

derekcollison: But flags override everything correct?

PR #97, opened 2019-06-07
Added a check to disable version checking if a local 0.0.0-dev version is the current version.

aricart:


main.go
```diff
@@ -17,18 +17,21 @@ func main() {
...
-		cmd.GetRootCmd().Printf("error checking for updates: %v\n", err)
-	}
-	if u != nil {
-		m, err := u.Run()
+	// disable update check if running a locally built version
+	if version != "0.0.0-dev" {
+		u, err := cmd.NewSelfUpdate()
 		if err != nil {
+			// not fatal
 			cmd.GetRootCmd().Printf("error checking for updates: %v\n", err)
```
sasbury: can it return after this?

aricart: no the execute is what initializes the tool.

cmd/update.go
```diff
@@ -126,6 +126,10 @@ func (u *SelfUpdate) Run() (*semver.Version, error) {
 }
 
 func (u *SelfUpdate) shouldCheck() bool {
+	have := semver.MustParse(GetRootCmd().Version).String()
+	if have == "0.0.0-dev" {
+		return false
```
derekcollison: I think this is good for now. However if I have a real old version we may want to be smarter.

Also, if I have not used the tool in a while and the first thing I see is it checking for an update when I just want to use the tool that is not a good user experience. We should be smarter about when we check IMO.

sasbury: checks are delayed 24hours i think so that should help (ie if you fail we don't check again for 24hours)

Could you write up a paragraph in an email or issue on what you think a good experience would be here?

derekcollison: Sent you a link.

aricart: maybe we can just find out what versions are out there, and if a help command/usage command kicks in, we could say version X - but version Y is available. The other thing would be to have a version file somewhere on github that we reference directly instead of letting the tool query for releases etc which likely takes much longer.

PR #100, opened 2019-06-20
Keyring

aricart:
This pull request changes the way that keys are stored and found.
Keys are now stored like: ~/.nsc/keys/<kind>/<shard>/<pubkey>.nk where the kind is the first letter of the public key describing its type O, A, U for operator, account and user. And a shard of the second and third letters in the public key. The key itself is stored using the full public key.

Credential files are stored as: ~/.nsc/creds/<operator>/<account>/<user>.creds

There's a migration step that will convert a keystore into the new layout. Old layout is preserved in a renamed directory.


cmd/accountcontextparams.go
```diff
@@ -32,18 +32,26 @@ func (p *AccountContextParams) BindFlags(cmd *cobra.Command) {
...
 		}
 		ctx.StoreCtx().Account.Name = p.Name
-		return nil
 	} else {
 		if config.Account != "" {
 			ctx.StoreCtx().Account.Name = config.Account
 			p.Name = config.Account
 		}
 	}
+	if p.Name != "" {
```
sasbury: why not combine with the earlier if?

aricart: the value tries to set from 2 different places. then that check, verifies it was set.

PR #106, opened 2019-06-25
FIX #103 - ~ in paths are not honored for signing keys.

### LINKED ISSUE #103: key resolution with ~/ paths fails

aricart:


cmd/signerparams_test.go
```diff
@@ -172,3 +173,30 @@ func Test_SignerParamsPathNotFound(t *testing.T) {
...
 }
+
+func Test_SignerParamsHomePath(t *testing.T) {
+	ts := NewTestStore(t, "O")
+	defer ts.Done(t)
+
+	home, err := homedir.Dir()
+	require.NoError(t, err)
+
+	// yeap not going to remove it
```
sasbury: well this is scary

aricart: just removing the file, not the dir :)

PR #109, opened 2019-06-25
modified signer params to:

aricart:
- present the user with a picker of all the keys we know we have
- if some keys are missing, provide an 'other' option
- validate that the entered key matches a signer public key

### DIFF
--- a/cmd/addimport_test.go
+++ b/cmd/addimport_test.go
@@ -119,7 +119,7 @@ func Test_AddImportInteractive(t *testing.T) {
 
 	cmd := createAddImportCmd()
 	HoistRootFlags(cmd)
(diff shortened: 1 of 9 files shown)


cmd/addimport_test.go
```diff
@@ -119,7 +119,7 @@ func Test_AddImportInteractive(t *testing.T) {
 
 	cmd := createAddImportCmd()
 	HoistRootFlags(cmd)
-	input := []interface{}{1, false, fp, "my import", "barfoo.>", ts.OperatorKeyPath}
+	input := []interface{}{1, false, fp, "my import", "barfoo.>", 0}
```
sasbury: do you have a test that doesn't use 0?

aricart: yes - the ones that provide a signing key directly.

PR #116, opened 2019-07-13
modified import process to:

aricart:
- enable user to pick from locally available exports
- if picked export is private, attempt to generate activation


cmd/addaccount_test.go
```diff
@@ -126,23 +126,30 @@ func Test_AddAccountInteractiveSigningKey(t *testing.T) {
 	defer ts.Done(t)
 
 	s1, pk1, _ := CreateOperatorKey(t)
-	_, pk2, _ := CreateOperatorKey(t)
 
-	_, _, err := ExecuteCmd(createEditOperatorCmd(), "--sk", pk1, "--sk", pk2)
+	t.Log("pk1", pk1)
```
derekcollison: You still want this here?

cmd/addaccount_test.go
```diff
@@ -126,23 +126,30 @@ func Test_AddAccountInteractiveSigningKey(t *testing.T) {
...
-
 	d, err := ts.Store.Read(store.JwtName("O"))
 	require.NoError(t, err)
 	oc, err := jwt.DecodeOperatorClaims(string(d))
 	require.NoError(t, err)
+
+	ac, err := ts.Store.ReadAccountClaim("A")
+	require.NoError(t, err)
+
+	t.Log(NewOperatorDescriber(*oc).Describe())
```
derekcollison: Same here

cmd/addimport.go
```diff
@@ -121,12 +121,147 @@ func (p *AddImportParams) SetDefaults(ctx ActionCtx) error {
...
+	}
+
+	ac, err := ctx.StoreCtx().Store.ReadAccountClaim(ctx.StoreCtx().Account.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	var filtered []AccountExport
+	for _, f := range found {
+		// FIXME: filtering on the target account, should eliminate exports the account already has
```
derekcollison: Not following this, what exactly are you filtering?

aricart: imports that match the exports they have

cmd/addaccount_test.go
aricart: thanks

cmd/addaccount_test.go
aricart: not really

cmd/addimport.go
derekcollison: Should we really filter that? Not sure the answer is obvious TBH.

PR #127, opened 2019-07-19
Added friendly names to account imports

aricart:
Describing an account shows import source accounts as simple names

### DIFF
--- a/cmd/describe.go
+++ b/cmd/describe.go
@@ -19,11 +19,47 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var WideFlag bool
+var Wide WideFun
+
+type WideFun = func(a string) string
+
+func noopNameFilter(a string) string {
+	return a
+}
+
+func friendlyNameFilter() (WideFun, error) {
+	m, err := friendlyNames(GetConfig().Operator)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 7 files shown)


cmd/describe.go
```diff
@@ -19,11 +19,47 @@ import (
...
+		return v
+	}, nil
+}
+
 var describeCmd = &cobra.Command{
 	Use:   "describe",
 	Short: "Describe assets such as operators, accounts, users, and jwt files",
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		var err error
+		if WideFlag {
```
derekcollison: Thought we got rid of wide flag?

cmd/describe.go
```diff
@@ -19,11 +19,47 @@ import (
...
+				return err
+			}
+		}
+		return nil
+	},
 }
 
 func init() {
 	GetRootCmd().AddCommand(describeCmd)
+	describeCmd.PersistentFlags().BoolVarP(&WideFlag, "long-ids", "W", false, "display account ids on imports")
```
derekcollison: Same.

cmd/describe.go
aricart: now it does showing the pk for the account in the imports.

PR #132, opened 2019-07-24
fix #128 moved creds parsing code to jwt lib

### LINKED ISSUE #128: nsc too brittle parsing creds files
We should use same regex the Go client uses. Simply adding a few extra dashes on the top banner line broke it.

aricart:


cmd/generatecreds.go
```diff
@@ -161,8 +161,7 @@ func GenerateConfig(s *store.Store, account string, user string, userKey nkeys.K
 		if err != nil {
 			return nil, fmt.Errorf("error getting seed: %v", err)
 		}
-
-		return FormatConfig("User", string(d), string(seed)), nil
+		return jwt.FormatUserConfig(string(d), seed)
```
sasbury: are there any other types of "config" i am ok if not, just curious

aricart: not yet

PR #139, opened 2019-07-26
fix #135 added deploy functionality

### LINKED ISSUE #135: deploy to operator

aricart:


cmd/common.go
```diff
@@ -361,3 +361,40 @@ func Expand(s string) (string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		body, _ := ioutil.ReadAll(resp.Body)
+		m := ""
+		if body != nil {
+			var vr jwt.ValidationResults
```
sasbury: This is the code i just got rid of in my PR :-) but yours will just go to the err and print the string so should be ok

aricart: I'll remove when I make the push cmd work the same.

PR #144, opened 2019-07-29
Added rtt tool

sasbury:
Added an rtt tool and updated help on req tool

### DIFF
--- a/cmd/rtttool.go
+++ b/cmd/rtttool.go
@@ -25,13 +25,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func createToolReqCmd() *cobra.Command {
-	var params ReqParams
+func createToolRTTCmd() *cobra.Command {
+	var params RttParams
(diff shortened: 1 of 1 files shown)


cmd/rtttool.go
```diff
@@ -0,0 +1,112 @@
...
+func (p *RttParams) PostInteractive(ctx ActionCtx) error {
+	return nil
+}
+
+func (p *RttParams) Validate(ctx ActionCtx) error {
+	if err := p.AccountUserContextParams.Validate(ctx); err != nil {
+		return err
+	}
+
+	if p.credsPath == "" {
```
aricart: This may be a bug - need to check if credsPath returns `""` when the fiel is not there. I believe it is just a path generation, so the file may not be there if you c/p the command likely this is a problem elsewhere.

aricart: actually the nats library will surface the error.

PR #147, opened 2019-07-31
Added support for setting max responses, and response expiry.

aricart:


cmd/adduser.go
```diff
@@ -133,6 +142,27 @@ func (p *AddUserParams) PreInteractive(ctx ActionCtx) error {
...
+	}
+	if ok {
+		p.respMax, err = cli.Prompt("Number of max responses", p.respMax, true, func(v string) error {
+			_, err := strconv.ParseInt(v, 10, 32)
+			if err != nil {
+				return err
+			}
+			return nil
+		})
+		p.respTTL, err = cli.Prompt("Response TTL duration", p.respTTL, true, func(v string) error {
```
derekcollison: It is the response permission ttl or expiration. Not sure if we want to reword here.

aricart: The field simply says how long the server will wait for the client to respond before removing the permission right? If that is the case, I wanted to avoid "Expiration" as we already use that name somewhere else, but really can be anything you like.

derekcollison: Yes maybe say Response Permission TTL?

PR #151, opened 2019-08-02
replaced init command with a simpler implementation

aricart:
new implementation only takes two options, 'name' and 'url', and creates an operator, account, and user using the specified name. If url is provided, it also deploys the created account to the specified url.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -19,10 +19,9 @@ import (
 	"fmt"
 	"path/filepath"
 
-	"github.com/nats-io/nsc/cli"
-
 	"github.com/nats-io/jwt"
 	"github.com/nats-io/nkeys"
+	"github.com/nats-io/nsc/cli"
 	"github.com/nats-io/nsc/cmd/store"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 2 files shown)


cmd/init.go
```diff
@@ -19,304 +19,243 @@ import (
...
-			if p.operator.create {
-				cmd.Println("Success! - initialized environment directory")
+			if params.Deploy == "" {
+				cmd.Printf("\n\nTo run a local server using this configuration, enter:\n")
+				cmd.Printf("> nsc generate config --mem-resolver --config-file <path/server.conf>\n")
+				cmd.Printf("start a nats-server using the generated config:\n")
+				cmd.Printf("> nats-server -c <path/server.conf>\n\n")
+
+				cmd.Printf("Or deploy your account to a managed service enter:\n")
+				cmd.Printf("> nsc deploy --url https://jwt.ngs.local:6060/jwt/v1/operator\n\n")
```
sasbury: worried about putting a real URL here, maybe just <operator url>

cmd/init.go
```diff
@@ -19,304 +19,243 @@ import (
...
-		p.operator.name = fmt.Sprintf("%s_operator", p.environmentName)
+func (p *EasyCmdParams) setOperatorDefaults(ctx ActionCtx) error {
+	oc, err := ctx.StoreCtx().Store.ReadOperatorClaim()
+	if err != nil {
+		return err
 	}
-
-	if p.account.name == "" {
-		p.account.name = fmt.Sprintf("%s_account", p.environmentName)
+	oc.OperatorServiceURLs.Add("nats://localhost:4222")
```
sasbury: should we tell them we did this somwehere?

cmd/init.go
```diff
@@ -19,304 +19,243 @@ import (
...
+				cmd.Printf("> nsc generate config --mem-resolver --config-file <path/server.conf>\n")
+				cmd.Printf("start a nats-server using the generated config:\n")
+				cmd.Printf("> nats-server -c <path/server.conf>\n\n")
+
+				cmd.Printf("Or deploy your account to a managed service enter:\n")
+				cmd.Printf("> nsc deploy --url https://jwt.ngs.local:6060/jwt/v1/operator\n\n")
 			}
 
+			cmd.Printf("To listen for messages enter:\n")
+			cmd.Printf("> nsc tools sub \">\"\n")
```
sasbury: maybe this is where we can say the operator JWT defaults to pointing at a local server

cmd/init.go
aricart: this is just a default value so that the:

nsc generate config ... 
stuff allows them to use the tools

aricart: they need to change it to their env whatever that is.

PR #156, opened 2019-08-09
Init cmd changes

aricart:
Support for local, remote, well known operator, with custom location for the project files.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -32,6 +33,7 @@ import (
 
 	"github.com/dustin/go-humanize"
 	"github.com/mitchellh/go-homedir"
+	"github.com/nats-io/jwt"
 	"github.com/nats-io/nkeys"
 	"github.com/nats-io/nsc/cli"
 	"github.com/nats-io/nsc/cmd/store"
@@ -315,6 +317,9 @@ func LoadFromURL(url string) ([]byte, error) {
 		return nil, fmt.Errorf("error loading %q: %v", url, err)
 	}
 	defer r.Body.Close()
+	if r.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("error reading response from %q: %v", url, r.Status)
+	}
 	var buf bytes.Buffer
 	_, err = io.Copy(&buf, r.Body)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


cmd/common.go
```diff
@@ -378,19 +383,64 @@ func Expand(s string) (string, error) {
...
+func AccountJwtURLFromString(asu string, accountSubject string) (string, error) {
+	u, err := url.Parse(asu)
+	if err != nil {
+		return "", err
+	}
+	u.Path = path.Join(u.Path, "accounts", accountSubject)
+	return u.String(), nil
+}
+
+func AccountJwtURL(oc *jwt.OperatorClaims, ac *jwt.AccountClaims) (string, error) {
```
sasbury: can you add doc to the JWT library for what you expect on these urls - also maybe prompt the correct form if someone is editing an operator jwt to have them

cmd/common.go
```diff
@@ -378,19 +383,64 @@ func Expand(s string) (string, error) {
...
+	}
+	return AccountJwtURLFromString(oc.AccountServerURL, ac.Subject)
+}
+
+func OperatorJwtURLFromString(asu string) (string, error) {
+	u, err := url.Parse(asu)
+	if err != nil {
+		return "", err
+	}
+	u.Path = path.Join(u.Path, "operator")
```
sasbury: synadia is currently special, but let's doc and I can add /operator to match /synadia in NGS.

cmd/contextconfig.go
```diff
@@ -32,16 +32,42 @@ type ContextConfig struct {
 }
 
 func NewContextConfig(storeRoot string) (*ContextConfig, error) {
```
sasbury: Step 2 is to make this a map with different roots for each operator, but reviewing step 1 only, wanted a comment to remind myself

cmd/deploy.go
```diff
@@ -292,5 +295,10 @@ func (p *DeployCmdParams) Run(ctx ActionCtx) error {
 		ad := NewAccountDescriber(*aac)
 		Write("--", []byte(ad.Describe()))
 	}
+
+	if IsAccountPending(status) {
```
sasbury: I don't use this yet, but now i will think about how i could use it :-)

cmd/wellknownoperator_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func Test_ListWellKnownOperators(t *testing.T) {
+	wko, err := GetWellKnownOperators()
+	require.NoError(t, err)
+	require.NotNil(t, wko)
+	require.True(t, len(wko) >= 1)
+
+	s, err := FindKnownOperator("SYNADIA")
```
sasbury: Maybe add checks for the URLs, even though it is a bit a==a it might be worth the safety in case we change to a text file or something.

cmd/common.go
aricart: OK we can change once that lands.

cmd/contextconfig.go
aricart: yes, we need to make this a map that is stored as a pref, then we can show the ones we know about.

cmd/wellknownoperator_test.go
aricart: not following

cmd/common.go
aricart: https://github.com/nats-io/jwt/pull/50

cmd/wellknownoperator_test.go
sasbury: I was just thinking there is value in having a double check that we set the URLs correctly in the code that specifies the known operators, so if that could changes we know we didn't break something

PR #157, opened 2019-08-13
support managed operators

aricart:
Typical operations in NSC require a signer that is locally available. To support a managed operator, the signer will never be available. Support for a managed operator is baked into the JWT via its Account Server URL field.

A managed operator accepts a self-signed account JWT on an account JWT server and re-signs it with the operator key.

The change for supporting a managed operator, the store function will check if the operator is managed, if it is, it will push the account JWT to the account server defined by the operator (a requirement for a managed operator). If pushed succeeds, it then queries the operator for the JWT which is now signed by the operator (with limits injected). The signed JWT is then stored on disk.

It is possible for an operator to accept an account (where additional steps may need to happen before the process is complete). At that point, a 'sync' command will pull the signed JWT at a later time in response to the user.

### DIFF
--- a/cmd/addaccount.go
+++ b/cmd/addaccount.go
@@ -20,9 +20,12 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/nats-io/jwt"
 	"github.com/nats-io/nkeys"
+	"github.com/nats-io/nsc/cli"
+	"github.com/nats-io/nsc/cmd/store"
 	"github.com/spf13/cobra"
 )
 
@@ -38,7 +41,7 @@ func CreateAddAccountCmd() *cobra.Command {
 			if err := RunAction(cmd, args, &params); err != nil {
 				return err
 			}
-			if params.generated && !QuietMode() {
+			if params.generate && !QuietMode() {
 				cmd.Printf("Generated account key - private key stored %q\n", AbbrevHomePaths(params.keyPath))
 			}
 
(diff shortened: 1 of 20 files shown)


cmd/addaccount.go
```diff
@@ -61,79 +64,184 @@ func init() {
...
+	if err != nil {
+		return nil, err
+	}
+	if t != nkeys.PrefixByteAccount {
+		return nil, errors.New("specified key is not a valid account nkey")
+	}
+	return nk, nil
+}
+
+func (p *AddAccountParams) validateAccountNKey(s string) error {
```
sasbury: is it worth the extra method? perhaps used as a validator in which case yes.

cmd/addaccount.go
```diff
@@ -61,79 +64,184 @@ func init() {
...
 		return err
 	}
 
 	return nil
 }
 
 func (p *AddAccountParams) Load(ctx ActionCtx) error {
+	var err error
+	if p.generate {
+		p.akp, err = nkeys.CreateAccount()
```
sasbury: maybe in the future (think make an issue) combine these two steps so that you either get a stored key or an error. Maybe not worth it just musing.

cmd/addaccount_test.go
```diff
@@ -111,13 +111,15 @@ func validateAddAccountClaims(t *testing.T, ts *TestStore) {
 	require.Equal(t, expire, ac.Expires)
 }
 
-func Test_AddAccountOperatorLessStore(t *testing.T) {
-	ts := NewTestStoreWithOperator(t, "test", nil)
+func Test_AddAccountManagedStore(t *testing.T) {
+	as, m := RunTestAccountServer(t)
+	defer as.Close()
+
+	ts := NewTestStoreWithOperatorJWT(t, string(m["operator"]))
```
sasbury: assuming this sets the "env"

cmd/push_test.go
```diff
@@ -86,114 +46,46 @@ func Test_SyncOK(t *testing.T) {
...
-	_, pk, kp := CreateOperatorKey(t)
-	oc := jwt.NewOperatorClaims(pk)
-	oc.Name = "O"
-	d, err := oc.Encode(kp)
-	require.NoError(t, err)
+	as, m := RunTestAccountServer(t)
+	defer as.Close()
 
-	ts := NewTestStoreWithOperatorJWT(t, d)
+	ts := NewTestStoreWithOperatorJWT(t, string(m["operator"]))
```
sasbury: i wonder if it will be more maintainable if this function just takes the map so you don't need the magic string everywhere

cmd/store/store.go
```diff
@@ -44,6 +49,93 @@ var standardDirs = []string{Accounts}
 
 var ErrNotExist = errors.New("resource does not exist")
 
+func OK(status int) bool {
+	return status == http.StatusOK
+}
+
+func Pending(status int) bool {
```
sasbury: I like these

cmd/store/store.go
```diff
@@ -321,19 +413,116 @@ func (s *Store) ListEntries(name ...string) ([]string, error) {
...
+		return nil, fmt.Errorf("invalid jwt: %v", err)
 	}
 	if gc.Name == "" {
-		return errors.New("jwt claim doesn't have a name")
+		return nil, errors.New("jwt claim doesn't have a name")
+	}
+	return &gc.Type, nil
+}
+
+func (s *Store) syncRemoteAccount(u string) (int, []byte, error) {
```
sasbury: i guess you have to take a URL here, but it feels weird that the store doesn't even know the account being sync'd as a result

cmd/store/store.go
```diff
@@ -321,19 +413,116 @@ func (s *Store) ListEntries(name ...string) ([]string, error) {
...
+	defer r.Body.Close()
+	var buf bytes.Buffer
+	_, err = io.Copy(&buf, r.Body)
+	if err != nil {
+		return 0, nil, fmt.Errorf("error reading response from %q: %v", u, err)
+	}
+	return r.StatusCode, buf.Bytes(), nil
+}
+
+func (s *Store) pushRemoteAccount(u string, data []byte) (int, []byte, error) {
```
sasbury: same comment on URLs

cmd/store/store.go
```diff
@@ -321,19 +413,116 @@ func (s *Store) ListEntries(name ...string) ([]string, error) {
...
+	resp, err := http.Post(u, "application/text", bytes.NewReader(data))
```
sasbury: you should use "application/jwt" here - to be pedantic

cmd/addaccount.go
aricart: Only add account and add operator need the key a priory because they may be generating the key that signs them.

cmd/push_test.go
aricart: it is a test, and the "operator" corresponds to id of the operator.

cmd/store/store.go
aricart: the coordination method extracts these, the others only do the needy http

cmd/push_test.go
sasbury: i know, just saying it is a "test account server" -> "test operator store" which i saw a bunch of places and has a magic string

PR #159, opened 2019-08-15
added check for a mixed keyring

aricart:
When using ngs and nsc, the keyring could get two different layouts


cmd/keys_test.go
```diff
@@ -40,6 +40,9 @@ func storeOldKey(ts *TestStore, operator string, account string, user string, se
 	// <op>/accounts/<actname>/<actname>.nk
 	// <op>/accounts/<actname>/users/<un>.creds
 	// <op>/accounts/<actname>/users/<un>.nk
+	// new key layout is:
+	// keys/<op>/<first letter pk>/<second and third letters pk>/<pk>.nk
```
derekcollison: I still think we should do away with this directory structure, overly complicates things IMO.

aricart: I agree.

PR #161, opened 2019-08-15
cwd awareness

aricart:
modified tool to be cwd aware
- if the cwd is a directory that immediately has an operator under it, the stores dir is set to it
- if the cwd is a sub directory belonging to an operator, the stores dir is set to the operator's parent dir
- anywhere else, uses the set preference

### DIFF
--- a/cmd/defaults.go
+++ b/cmd/defaults.go
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/nats-io/jwt"
+
 	"github.com/mitchellh/go-homedir"
 	"github.com/nats-io/nsc/cmd/store"
 )
@@ -70,7 +72,7 @@ func GetCwdCtx() *ContextConfig {
 	}
 	if ok {
 		ctx.StoreRoot = filepath.Dir(dir)
-		ctx.Operator = info.EnvironmentName
+		ctx.Operator = info.Name
 		return &ctx
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/defaults.go
```diff
@@ -54,6 +57,84 @@ func GetConfig() *ToolConfig {
...
+	ok, err := isOperatorDir(dir)
+	if err != nil {
+		return ""
+	}
+	if ok {
+		return filepath.Dir(dir)
+	}
+
+	// search down
+	infos, err := ioutil.ReadDir(dir)
```
sasbury: This is scary, what if there is more than one down there, you just take the first one? I was thinking we only go up

cmd/defaults.go
```diff
@@ -88,7 +168,14 @@ func LoadOrInit(github string, toolHomeEnvName string) (*ToolConfig, error) {
 		if err := config.Save(); err != nil {
 			return nil, err
 		}
+	} else {
```
sasbury: not sure how to do this with NGS, right now i just use tool home, not a comment for this code, but seeing what you think

cmd/defaults.go
aricart: all we are determining here is there's an operator directory right below.
The code for selection is:
```	switch len(choices) {
	case 0:
		return ""
	case 1:
		return choices[0]
	default:
		for _, v := range choices {
			if v == current {
				return current
			}
		}
		return ""
	}```

aricart: If there's one, that it is selected, otherwise, it is only set if it matches a value calculated.

PR #178, opened 2019-09-06
use correct import path for survey

ripienaar:
The import path for survey should be `gopkg.in/AlecAivazis/survey.v1` but
`github.com/AlecAivazis/survey` was used in this project, anyone who uses
`go mod` to load anything from this project would end up needing both 
paths in their dependencies as the module internally use the gopkg one.


ripienaar:
I am not sure why my mod vendor deleted the `olang.org/x/sys/unix` and `golang.org/x/text/unicode` stuff though, perhaps because I am on go 1.13 and some of these are now in core?

aricart:
I was about to upgrade to v2 today.

ripienaar:
awesome, I noticed v2 after this PR and was going to do it for you :) So if you'll fix the path when you do that we can kill this PR

aricart:
Sounds good. There's a few changes I have to make.

ripienaar:
@aricart any update on this v2 business?

aricart:
not yet - started work on it, but will be refactoring all the
cli widgets along with that change.

aricart:
This is fixed by #242.

PR #179, opened 2019-09-06
Self-sign migrate

aricart:
The migrate command needs to self-sign accounts before pushing to the remote server, or the account server will reject them.


cmd/migrate.go
```diff
@@ -24,6 +24,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/nats-io/nkeys"
+
```
kozlovic: don't need new line here, which should then have it alpha ordered with the rest of the "github" imports.

aricart: fixed

PR #183, opened 2019-09-06
FIX #182 - Modified add operator to allow the user to enter a well-known name.

### LINKED ISSUE #182: add operator using well known name
Well known operators are baked into the app or set in the env via NSC_name_OPERATOR.
To simplify the setup for managed operators like ngs, using a well know operator name saves the user from typing a long URL.

Instead of specifying a file/url allow the url field to have the 'name' of the well known operator.

aricart:


cmd/addoperator.go
```diff
@@ -87,7 +88,21 @@ func (p *AddOperatorParams) PreInteractive(ctx ActionCtx) error {
...
-		p.jwtPath, err = cli.Prompt("path or url for operator jwt", p.jwtPath, true, cli.PathOrURLValidator())
+		p.jwtPath, err = cli.Prompt("path or url for operator jwt", p.jwtPath, true, func(v string) error {
+			// is it is an URL or path
+			pv := cli.PathOrURLValidator()
+			if err := pv(v); err != nil {
+				// if it doesn't exist - could it be the name of well known operator
+				if os.IsNotExist(err) {
+					wko, _ := FindKnownOperator(v)
+					if wko != nil {
+						return nil
```
kozlovic: This does not make sense to me. If `wko` is not nil, then we return `nil` (for no error), but do the same if it is `nil`..

cmd/addoperator.go
```diff
@@ -87,7 +88,21 @@ func (p *AddOperatorParams) PreInteractive(ctx ActionCtx) error {
...
+					}
+					return nil
+				}
+			}
+			return nil
```
kozlovic: So we never return an error in this validation function..

cmd/addoperator.go
aricart: it is a validator - there's no error in that case.

kozlovic: Ok, but look at the code, this does not make sense. You return the same result (nil) regardless if wko is nil or not.

cmd/addoperator.go
aricart: yes this last return is wrong, it needs to return err at that point.

cmd/addoperator.go
```diff
@@ -87,8 +88,20 @@ func (p *AddOperatorParams) PreInteractive(ctx ActionCtx) error {
...
+			if err := pv(v); err != nil {
+				// if it doesn't exist - could it be the name of well known operator
+				if os.IsNotExist(err) {
+					wko, _ := FindKnownOperator(v)
+					if wko != nil {
+						return nil
+					}
+				}
+			}
+			return err
```
kozlovic: This is wrong too. Here should be probably return `nil`, but you should have a `return err` at line 101. That is, return the error from pv(v). Currently you are returning `err` that is defined before, say line 91.

PR #186, opened 2019-09-11
Misc changes from sasbury

sasbury:
Three change lists for 3 changes:

1- Added support for random names in init, add user and add account
2- Added --nats flag for nats URL in tools
3 - Updated expiration help string


sasbury:
I tested manually, not sure of the best way to do a unit tests for the random names

PR #202, opened 2019-09-18
Collapsing status hid server messages. Now if a child status contains

aricart:
server messages, those messages are shown.

### DIFF
--- a/cmd/store/status.go
+++ b/cmd/store/status.go
@@ -199,8 +199,7 @@ func (r *Report) printsDetails() bool {
 
 func (r *Report) HasServerMessages() bool {
 	for _, v := range r.Details {
-		_, ok := v.(*ServerMessage)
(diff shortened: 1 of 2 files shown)


cmd/store/status.go
```diff
@@ -192,7 +192,17 @@ func (r *Report) printsDetails() bool {
...
 	case DetailsOnErrorOrWarning:
-		return r.StatusCode != OK
+		return r.StatusCode != OK || r.HasServerMessages()
+	}
+	return false
+}
+
+func (r *Report) HasServerMessages() bool {
+	for _, v := range r.Details {
+		_, ok := v.(*ServerMessage)
```
kozlovic: You can make this a one liner:
```
if _, ok := v.(*ServerMessage); ok {
    return true
}
```

PR #207, opened 2019-09-19
Removed defaults assigned to the response permissions.

aricart:


kozlovic:
Let's wait on this one until we have the server setting the defaults, otherwise, not sure things will behave well with 0 values...

aricart:
edit user shared a lot of the code and was not dry

cmd/adduser.go
```diff
@@ -311,3 +266,98 @@ func (p *AddUserParams) editUserClaim(c interface{}, ctx ActionCtx) error {
...
 	return nil
 }
+
+type ResponsePermsParams struct {
+	respTTL string
+	respMax string
+	rmResp  bool
+}
+
+func (p *ResponsePermsParams) maxResponseValidator(s string) error {
```
kozlovic: Just checked with Derek and values `<0` should be rejected (0 is ok, server will use default).

cmd/adduser.go
```diff
@@ -311,3 +266,98 @@ func (p *AddUserParams) editUserClaim(c interface{}, ctx ActionCtx) error {
...
+		return nil
+	}
+	_, err := strconv.Atoi(s)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (p *ResponsePermsParams) ttlValidator(s string) error {
```
kozlovic: Same than for max msgs.

cmd/adduser.go
kozlovic: Retracting that statement, will use negative in the server to make the value "infinite"

cmd/adduser.go
```diff
@@ -311,3 +266,108 @@ func (p *AddUserParams) editUserClaim(c interface{}, ctx ActionCtx) error {
...
+func (p *ResponsePermsParams) maxResponseValidator(s string) error {
+	_, err := p.parseMaxResponse(s)
+	return err
+}
+
+func (p *ResponsePermsParams) parseMaxResponse(s string) (int, error) {
+	if s == "" {
+		return 0, nil
+	}
+	v, err := strconv.Atoi(s)
```
kozlovic: Feels that this all could be `return strconv.Atoi(s)`(assuming that caller does not do anything with `v` if `err !=nil`.

cmd/adduser.go
```diff
@@ -311,3 +266,108 @@ func (p *AddUserParams) editUserClaim(c interface{}, ctx ActionCtx) error {
...
+func (p *ResponsePermsParams) ttlValidator(s string) error {
+	_, err := p.parseTTL(s)
+	return err
+}
+
+func (p *ResponsePermsParams) parseTTL(s string) (time.Duration, error) {
+	if s == "" {
+		return time.Duration(0), nil
+	}
+	d, err := time.ParseDuration(s)
```
kozlovic: Probably same, likely `return time.ParseDuration(s)` here would do the same (did not check what ParseDuration returns as duration in case of error, but again, likely caller is not going to use the duration if error returned).

cmd/adduser.go
aricart: doesn't - if validation fails, it's done

cmd/adduser.go
aricart: if validation fails, the whole thing stops at that point - unless you are interactively editing. in which case you continue to try. After that, the value is never going to be an error as it already passed, but I need to handle it or it is not obvious. IE - if there's an error it won't ever record it

PR #212, opened 2019-09-20
changes to normalize and make it easier to select imports/exports when deleting them

aricart:
- on interactive streams are rendered with `-> subject` and services with `<- subject`
- add import flag `--local-subject` changed description as 'local subject or prefix'
- stream imports no longer force a prefix
- removed `--service` flag from delete import - the unique identifier is the remote subject
- for non interactive imports delete, if the subject has multiple matches, `--src-account` with account key is required

### DIFF
--- a/cmd/addimport.go
+++ b/cmd/addimport.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 The NATS Authors
+ * Copyright 2018-2019 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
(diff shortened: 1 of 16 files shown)


cmd/addimport.go
```diff
@@ -391,9 +393,30 @@ func (p *AddImportParams) initFromActivation(ctx ActionCtx) error {
 	return nil
 }
 
+func (p *AddImportParams) checkServiceSubject(s string) error {
```
sasbury: Do we need to check for streams, with the same subject and no prefix

sasbury: talked in person, ok, not wrong but maybe a warning

PR #230, opened 2019-11-15
encrypt req/sub for kv demo

aricart:
internal -E and -D flags for nsc req to encrypt requests and decrypt responses.
internal flag -D for nsc sub to decrypt messages


cmd/tool.go
```diff
@@ -63,3 +74,100 @@ func createDefaultToolOptions(name string, ctx ActionCtx) []nats.Option {
...
+			decoded = append(decoded, d)
+		}
+		return bytes.Join(decoded, []byte(" ")), nil
+	} else {
+		return decrypt(pk, data)
+	}
+}
+
+func decrypt(pk string, data []byte) ([]byte, error) {
+	var b32Enc = base32.StdEncoding.WithPadding(base32.NoPadding)
```
derekcollison: base32 will blow up the data pretty good, would use base64 at minimum.

cmd/tool.go
```diff
@@ -63,3 +74,100 @@ func createDefaultToolOptions(name string, ctx ActionCtx) []nats.Option {
...
+	// seed the nonce with the same seed so that we have predictable encryption
+	if _, err = io.ReadFull(strings.NewReader(pk), nonce); err != nil {
+		return nil, fmt.Errorf("error generating random sequence: %v", err)
+	}
+	// encrypt the data
+	raw := g.Seal(nonce, nonce, data, nil)
+
+	// encode the data
+	var b32Enc = base32.StdEncoding.WithPadding(base32.NoPadding)
+	buf := make([]byte, b32Enc.EncodedLen(len(raw)))
```
derekcollison: See below, blows up data too much IMO. base64.

cmd/reqtool.go
```diff
@@ -117,8 +129,23 @@ func (p *ReqParams) Run(ctx ActionCtx) (store.Status, error) {
 		payload = ctx.Args()[1]
 	}
 
-	ctx.CurrentCmd().Printf("published request: [%s] : '%s'\n", subj, payload)
-	msg, err := nc.Request(subj, []byte(payload), 5*time.Second)
+	var seed string
+	if encryptFlag || decryptFlag {
```
derekcollison: Since we are doing symmetric for now, just one flag?

cmd/tool.go
```diff
@@ -63,3 +74,100 @@ func createDefaultToolOptions(name string, ctx ActionCtx) []nats.Option {
 	}))
 	return opts
 }
+
+func createCypher(pk string) (cipher.AEAD, error) {
+	// hash the provided private nkey into 32 bytes
+	hash := sha256.Sum256([]byte(pk))
+	c, err := aes.NewCipher(hash[0:32])
```
derekcollison: Don't need 0

PR #233, opened 2019-11-20
Command to fix an env

aricart:


cmd/fixenv.go
```diff
@@ -0,0 +1,700 @@
+/*
+ * Copyright 2018-2019 The NATS Authors
```
kozlovic: Date should start at 2019

cmd/fixenv_test.go
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright 2018-2019 The NATS Authors
```
kozlovic: Date should start at 2019

cmd/fixenv.go
```diff
@@ -0,0 +1,700 @@
...
+	ot.ActToUsers = make(map[string]jwt.StringList)
+	return &ot
+}
+
+func createFixCmd() *cobra.Command {
+	var params FixCmd
+
+	var cmd = &cobra.Command{
+		Use:   "fix",
+		Short: "rebuilds a project tree from screwed the contents of one or more directories",
```
kozlovic: Don't think there should be `screwed` in the description..maybe: "from corrupted contents on one or more directories"?

cmd/fixenv.go
```diff
@@ -0,0 +1,700 @@
...
+
+	if err := p.Regenerate(rr); err != nil {
+		return rr, err
+	}
+
+	return rr, nil
+}
+
+func (p *FixCmd) Regenerate(rr *store.Report) error {
+	if p.out == "" {
```
kozlovic: So this will create a "fix_" directory.. Should it be deleted in case of failure? And on success, should that be renamed? How user would use this "fix"?

cmd/fixenv.go
aricart: lol yes

cmd/fixenv_test.go
aricart: fixed

cmd/fixenv.go
aricart: fixed

cmd/fixenv.go
aricart: The tool will actually print what you have to do, I also added an example section - note the tool is hidden and won't show for anyone unless they know it is there.

The only place where it can fail is on reading files, and creating the output directories. If it fails, you should be able to repeat the imports by adding new directories along with the 'fixed' one

Things that will fail are:
- no operator is found
- account references an unknown operator
- user references an unknown account

There's also an option to load cred files, which will read the cred files and import jwts and nkeys from them.

PR #234, opened 2019-11-21
implemented delete user

### LINKED ISSUE #228: delete or revoke users
add command to delete/revoke users

aricart:
FIX #228


aricart:
> Question: the invalid flags will be caught - I had the intention of allowing you to do without the `-n` flag, but didn't check it.

PR #242, opened 2019-12-03
moved to cliprompts/v2

aricart:
note that for some reason one of the tests triggered a failure as the jwt library was returning an error on a metric rate of `0` which is valid (to disable the metric). This is currently referencing the jwt branch with the fix.


aricart:
This fixes #178

PR #244, opened 2019-12-31
Added delete account. Deleting an account also deletes its users.

aricart:
Options are provided to revoke and delete nkeys and creds files for the entities. When an account is deleted, the JWT is set to expire 1m after. This allows an account server to reject any other user in the wild.

### DIFF
--- a/cmd/deleteaccount.go
+++ b/cmd/deleteaccount.go
@@ -1,0 +1,264 @@
+/*
+ * Copyright 2019 The NATS Authors
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"time"
+
+	cli "github.com/nats-io/cliprompts/v2"
+	"github.com/nats-io/jwt"
+	"github.com/nats-io/nkeys"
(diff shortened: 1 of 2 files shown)


cmd/deleteaccount.go
```diff
@@ -0,0 +1,264 @@
...
+
+	return nil
+}
+
+func (p *DeleteAccountParams) PreInteractive(ctx ActionCtx) error {
+	var err error
+	if err = p.AccountContextParams.Edit(ctx); err != nil {
+		return err
+	}
+	p.revoke, err = cli.Confirm("revoke all account users before deleting", true)
```
derekcollison: When we revoke users we formally need the account to still be active to hold the revocations. Should this be named slightly different maybe?

kthomas:
"Deleting" an account is not meant to be compatible with our contemplated bearer authorization, right?

cmd/deleteaccount.go
aricart: The delete account process does the following:
- Revokes all users
- Removes all account signing keys
- Sets the expiration of the account JWT to 1m into the future
- Pushes the account JWT to the account server (if any is specified in by the operator)
- Deletes the user JWTs
- Deletes the local account JWT
- Deletes user creds if specified
- Deletes the user, account signing keys, and account nkeys if specified

The remote system will have a record, but any user signed account will not be valid. If the nkeys to the account are preserved, it may be possible to re-issue (even sync the remote one), and start over.

aricart: > "Deleting" an account is not meant to be compatible with our contemplated bearer authorization, right?

Deleting an account means invalidating an entity that issues users. If you have a bearer token (an user) that was issued by an account and that account is not valid in the system, the user is also not valid.

derekcollison: Agree, should we have the operator revoke the account?

kthomas: I see... this essentially revokes the key which would be referenced by `kid` in the bearer token header.

The ability to revoke a key obviously makes sense, but under the bearer model it seems odd that a user's JWTs exist on the server and can be "deleted."

derekcollison: Server never holds onto a user jwt, so no worries there.

PR #248, opened 2020-01-14
Added support for editing bearer token field in users

aricart:
This PR currently references a nats-io/jwt branch  - before this is merged we need to update to a released JWT.


cmd/edituser_test.go
```diff
@@ -303,3 +303,28 @@ func Test_EditUserResponsePermissions(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Contains(t, stderr, "changed bearer to true")
+
+	u, err = ts.Store.ReadUserClaim("A", "U")
+	require.NoError(t, err)
+	require.True(t, u.BearerToken)
+
+	_, stderr, err = ExecuteCmd(createEditUserCmd(), "--name", "U", "--bearer=false")
+	require.NoError(t, err)
+	require.Contains(t, stderr, "changed bearer to false")
```
kozlovic: should you read the claim again and test that it has u.BearerToken is now false?

aricart: yes

PR #253, opened 2020-01-29
Added some usage examples documenting use of --max-responses.

aricart:
Simplified --max-responses so that the cli itself attempts the validation of the flag as a number.


aricart:
Figured out how to allow the usages of a bare flag (with a default of `1`) or with an arbitrary value.

This means that we can do `--allow-pub-response` and `--allow-pub-response=10`.

cmd/adduser.go
```diff
@@ -37,7 +37,32 @@ func CreateAddUserCmd() *cobra.Command {
...
+nsc add user --name <n> --allow-pubsub <subject>,...
+nsc add user --name <n> --allow-pub <subject>,...
+nsc add user --name <n> --allow-sub <subject>,...
+
+# Set permissions so that the user cannot publish nor subscribe to the specified subjects or wildcards:
+nsc add user --name <n> --deny-pubsub <subject>,...
+nsc add user --name <n> --deny-pub <subject>,...
+nsc add user --name <n> --deny-sub <subject>,...
+
+# To dynamically allow publishing to reply subjects
```
derekcollison: To dynamically allow publishing to reply subjects, this works well for service responders.

aricart: Fixed!

PR #257, opened 2020-02-12
Added one more Execute with Writer decided by the caller function

shantanubansal:


shantanubansal:
Will be very helpful if you guys can please review and approve the request asap

shantanubansal:
@aricart Here test case coverage won't affect anything. Please if you review the change

aricart:
@shantanubansal note that if you are embedding nsc, nsc is not an api. It is a tool, the api for generating the JWTs is github.com/nats-io/jwt and github.com/nats-io/nkeys

shantanubansal:
@aricart 
i have a suggestion. If we can have HTTP client for NSC. That will be of much better use plus it can have nats-account-server implemented within it

PR #258, opened 2020-02-14
Add a portable shell install script

philpennock:
This is not quite POSIX shell, we do ask for `local` to be available, and a few basic commands.  It's about as portable as I can make it without twisting into deep knots which risk adding bugs.

External commands required beyond POSIX: curl unzip mktemp

Document the usage of this script.

Adjust the Python to also use ~/.bashrc not ~/.bash_profile because the profile won't be run in all situations while the rc file should be used whenever the user has an interactive bash shell.

Usage for the shell differs slightly from the Python; it has a `-h` help option:

    % ./install.sh -h
    Usage: install [-t <tag>] [-d <dir>] [-s <dir>]
     -d dir     directory to download into [default: ~/.nsccli/bin]
     -s dir     directory in which to place a symlink to the binary
                [default: ~/bin] [use '-' to forcibly not place a symlink]
     -t tag     retrieve a tagged release instead of the latest

### DIFF
--- a/install.py
+++ b/install.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright 2018 The NATS Authors
+# Copyright 2018-2020 The NATS Authors
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -31,8 +31,6 @@ except ImportError:
     from urllib2 import urlopen
 
 NSC_REPO_URL = "https://github.com/nats-io/nsc"
-LATEST_RELEASE_URL = NSC_REPO_URL + "/releases/latest"
-TAG_URL = NSC_REPO_URL + "/releases/tag/"
 FILENAME_LOOKUP = {
     "darwin": "nsc-darwin-amd64.zip",
     "linux": "nsc-linux-amd64.zip",
@@ -50,22 +48,10 @@ def release_url(platform, tag):
         print("Unable to locate appropriate filename for", platform)
         sys.exit(1)
 
-    url = TAG_URL + tag if tag else LATEST_RELEASE_URL
-
-    try:
-        html = urlopen(url).read().decode('utf-8')
-    except:
-        print("Unable to find release page for", tag)
-        sys.exit(1)
-
-    urls = re.findall(r'href=[\'"]?([^\'" >]+)', html)
-    matching = [u for u in urls if filename in u]
-
-    if len(matching) != 1:
-        print("Unable to find download url for", filename)
-        sys.exit(1)
-
-    return "https://github.com" + matching[0]
+    release_base = NSC_REPO_URL + '/releases/'
+    if tag:
+        return release_base + 'download/' + tag + '/' + filename
+    return release_base + 'latest/download/' + filename
 
 
 def download_with_progress(url):
@@ -116,6 +102,9 @@ def main():
     os.chmod(exe_fn, 0o744)
 
     print("NSC: " + exe_fn)
+    if maybe_symlink(exe_fn):
+        return
+
     print("Now manually add %s to your $PATH" % bin_dir)
     if windows:
         print("Windows Cmd Prompt Example:")
@@ -123,13 +112,49 @@ def main():
         print()
     else:
         print("Bash Example:")
-        print("  echo 'export PATH=\"$PATH:%s\"' >> $HOME/.bash_profile" % bin_dir)
-        print("  source $HOME/.bash_profile")
+        print("  echo 'export PATH=\"$PATH:%s\"' >> $HOME/.bashrc" % bin_dir)
+        print("  source $HOME/.bashrc")
         print()
         print("Zsh Example:")
         print("  echo 'export PATH=\"$PATH:%s\"' >> $HOME/.zshrc" % bin_dir)
         print("  source $HOME/.zshrc")
         print()
+
+
+# Returns True if install instructions are not needed
+def maybe_symlink(exe_fn):
+    sym_dir = nsc_symlink_dir()
+    if not sym_dir:
+        return False
+    link_path = os.path.join(sym_dir, os.path.basename(exe_fn))
+    if os.path.exists(link_path):
+        if os.path.islink(link_path):
+            try:
+                os.unlink(link_path)
+            except Exception:
+                return False
+        else:
+            print("Not touching non-symlink: " + link_path)
+            return False
+    try:
+        os.symlink(exe_fn, link_path)
+        print("NSC: " + link_path)
+        if dir_in_PATH(sym_dir):
+            return True
+        return False
+    except Exception:
+        # Python2 does not support symlinks on Windows, amongst other
+        # reasons this might have failed.
+        return False
+
+
+def dir_in_PATH(dirname):
+    try:
+        envPath = os.environ['PATH']
+    except Exception:
+        return False
+    return dirname in envPath.split(os.pathsep)
+
 
 def mkdir(d):
     if not os.path.exists(d):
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -0,0 +1,229 @@
...
+# We are sh, not bash; we might want bash/zsh for associative arrays but some
+# OSes are currently on bash3 and removing bash, while we don't want a zsh
+# dependency; so we're sticking to "pretty portable shell" even if it's a
+# little more convoluted as a result.
+#
+# We rely upon the following beyond basic POSIX shell:
+#  1. A  `local`  command (built-in to shell; almost all sh does have this)
+#  2. A  `curl`   command (to download files)
+#  3. An `unzip`  command (to extract content from .zip files)
+#  4. A  `jq`     command (to manipulate JSON)
```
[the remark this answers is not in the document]
philpennock: You make a fair point and in the past I've used the same sorts of patterns.  I'm not aware of them being guaranteed, whereas this approach strictly follows GitHub's APIs.  I'm willing to switch.  Whereabouts is the documentation of the stability of these hostname patterns?  (Can switch without those, but would be nice to comment in the script to justify the fixed nature, especially when those have changed at least once in the past.)

Pragmatically, jq is default in most environments I've seen which are complete enough to claim POSIX, and it's de facto standard.  I was writing to match the Python closely, but didn't want to parse HTML in the same way.  This seemed like a more solid approach.

But if we have a guarantee from GitHub, or are willing to just rewrite should it prove necessary, then yes we can drop this.  We can also then simplify the Python version too.

install.sh
```diff
@@ -0,0 +1,229 @@
...
+readonly RELEASE_ARCH="amd64"
+
+# Finding the releases to download
+readonly GITHUB_OWNER_REPO='nats-io/nsc'
+readonly GITHUB_API_NSC="https://api.github.com/repos/$GITHUB_OWNER_REPO"
+readonly HTTP_USER_AGENT='nsc_install/0.1 (@nats-io)'
+readonly GITHUB_HTTP_ACCEPT='Accept: application/vnd.github.v3+json, application/json'
+
+# Where to install to, relative to home-dir
+readonly NSC_RELATIVE_BIN_DIR='.nsccli/bin'
```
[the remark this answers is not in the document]
philpennock: My best guess is that it's because the Python also runs on Windows, where we don't have such niceties, so picking one directory and sticking with it, cross platform, is the way forward.

I don't think we want to make the install location vary by which interpreter was chosen for the install script, so we absolutely need to match what the Python program does.  We change both or neither.

Sticking to the current location but adding a symlink in ~/bin if ~/bin exists and is in $PATH seems sane.

install.sh
```diff
@@ -0,0 +1,229 @@
...
+
+  note "Extracting ${want_filename} from $stage_zipfile"
+  # But unzip(1) does not let us override permissions and it does not obey
+  # umask so the file might now exist with overly broad permissions, depending
+  # upon whether or not the local environment has a per-user group which was
+  # used.  We don't know that the extracting user wants everyone else in their
+  # current group to be able to write to the file.
+  # So: extract into the temporary directory, which we've forced via umask to
+  # be self-only, chmod while it's safe in there, and then move it into place.
+  #   -b forces binary
```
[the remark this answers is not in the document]
philpennock: Oh crap.  Well, unzip should unzip the file safely without it.  I'm glad I documented it so explicitly.  Thanks.

install.sh
```diff
@@ -0,0 +1,229 @@
...
+normalized_ostype() {
+  local ostype
+  # We only need to worry about ASCII here
+  ostype="$(uname -s | tr A-Z a-z)"
+  case "$ostype" in
+    (*linux*)  ostype="linux" ;;
+    (win32)    ostype="windows" ;;
+    (ming*_nt) ostype="windows" ;;
+  esac
+  printf '%s\n' "$ostype"
```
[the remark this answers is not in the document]
philpennock: That depends upon the contents of `$ostype` and whether it holds any backslashes, and then which version of echo(1) is in play.  printf(1) is more standard than echo: it's in POSIX and its behavior doesn't vary (at this level) by shell.  So this construct works in sh/ash/ksh/bash/zsh across all historical Unices for "a few decades" no matter what's in `$ostype`; whereas echo is more problematic.

install.sh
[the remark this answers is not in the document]
philpennock: I found documentation for one, and just rolled with it for the other.

install.sh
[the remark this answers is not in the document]
philpennock: I haven't changed the Python yet.  I'm creating the symlink if `~/bin/` exists, and listing it, then constraining the installation instructions based upon that being in $PATH.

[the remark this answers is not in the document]
philpennock: Python updated.

install.sh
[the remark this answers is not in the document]
ripienaar: hmm, from posix docs this is exactly what echo is supposed to do, anyway, same same - you use this consistently so thats best.  I was just curious what the specific reason was, all good

install.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright 2018 The NATS Authors
+# Copyright 2018,2020 The NATS Authors
```
derekcollison: 2018-2020 is style we have been using.

install.sh
```diff
@@ -33,9 +45,7 @@ readonly RELEASE_ARCH="amd64"
 
 # Finding the releases to download
 readonly GITHUB_OWNER_REPO='nats-io/nsc'
-readonly GITHUB_API_NSC="https://api.github.com/repos/$GITHUB_OWNER_REPO"
 readonly HTTP_USER_AGENT='nsc_install/0.1 (@nats-io)'
-readonly GITHUB_HTTP_ACCEPT='Accept: application/vnd.github.v3+json, application/json'
 
 # Where to install to, relative to home-dir
 readonly NSC_RELATIVE_BIN_DIR='.nsccli/bin'
```
derekcollison: This always been the case?

philpennock: Has what always been the case?  The logic was there for talking to GitHub's API server and there's a bunch of features which only turn on with a correct `Accept:` header ... I forget the details, I ripped the string from my own `github_curl` wrapper.

When we're not talking to the API server, we don't need this constant.

install.py
philpennock: fixed

derekcollison: I still see the comma version, wanted 2018-2020.

install.sh
derekcollison: I meant this.. '.nsccli/bin'

philpennock: Oh.  Yes, that's where the Python script installs to.  I duplicated the functionality closely.

`git annotate` shows those lines as coming from commit 5c41b492 which is the commit which introduced the Python script, so "yes, always".

philpennock:
nb: the build failure after the force-push was because the branch was deleted before Travis could get the content.  The force push was of the squashed commit.

philpennock:
```
% git rev-parse '116bfd8^{tree}'  
21d2cdfcc5eb9582a781648e6f2ce0c48017d9a4
% git rev-parse 'fb12d6b^{tree}'            
21d2cdfcc5eb9582a781648e6f2ce0c48017d9a4
```

install.sh
aricart: ??? if you run the python script it installs it in ~/.nsc, not ~/.nsccli

aricart: Not quite sure how this came to be. - Indeed it does store it in .nsccli, really we should normalize this stuff, and store it in .nats

install.py
philpennock: You might have been looking at an outdated version, per GitHub's yellow "Outdated" boxes.  The code as landed on master is very clearly `2018-2020`.

derekcollison: ok thx

Issue #259, opened 2020-02-20
nsc primarily writes to stderr


matthiashanel:
only nsc describe writes to stdout.

PR #260, opened 2020-02-20
[FIXED] #259 Write to stdout instead of stderr

matthiashanel:
setting the output on cli alone was not enough.


matthiashanel:
@aricart can we have a look at this again, been asked about this on slack.

aricart:
seems to be working correctly

PR #262, opened 2020-03-11
windows issues

### LINKED ISSUE #261: nsc under windows issues
- loading resources from paths/or urls generically, fails as the code simply checks if there's a scheme. In windows, a plain file is parsed as an URL with a scheme matching the drive letter.
- updating a remote account server fails, as the URL path in one condition is built using `filepath.Join` instead of `path.Join`.
- printing paths double prints the backslashes.

aricart:
FIX #261 nsc windows issues
- loading resources from paths/or urls generically, fails as the code simply checks if there's a scheme. In windows, a plain file is parsed as an URL with a scheme matching the drive letter.
- updating a remote account server fails, as the URL path in one condition is built using `filepath.Join` instead of `path.Join`.
- printing paths double prints the backslashes.


.travis.yml
```diff
@@ -15,14 +19,13 @@ install:
...
-#- misspell -error -locale US .
-#- megacheck $EXCLUDE_VENDOR
+  - |
+    if [ "$TRAVIS_OS_NAME" == "linux" ];
+    then EXCLUDE_VENDOR=$(go list ./... | grep -v "/vendor/") && $(exit $(go fmt $EXCLUDE_VENDOR | wc -l)) && go vet $EXCLUDE_VENDOR
+    fi
 
 script:
-- ./scripts/cov.sh TRAVIS
+    go test -coverpkg=./... -coverprofile=./coverage.out ./...
```
ColinSullivan1: can cov.sh be removed now?

aricart: yea I think it could be

cmd/push_test.go
```diff
@@ -74,7 +75,11 @@ func Test_SyncNoServer(t *testing.T) {
 
 	_, stderr, err := ExecuteCmd(createPushCmd(), "--account", "A")
 	require.Error(t, err)
-	require.Contains(t, stderr, "connect: connection refused")
+	if runtime.GOOS == "windows" {
+		require.Contains(t, stderr, "connectex: No connection")
```
kozlovic: connectex?

cmd/common.go
```diff
@@ -291,12 +291,16 @@ func SeedNKeyValidatorMatching(kind nkeys.PrefixByte, pukeys []string) cli.Valid
 
 func IsURL(v string) bool {
 	u, err := url.Parse(v)
```
ColinSullivan1: ...could be one if statement.

.travis.yml
aricart: fixed

cmd/common.go
aricart: fixed

cmd/push_test.go
aricart: That is what windows 10 reports...

aricart:
> LGTM. Not sure why you replaced all %q with %#q especially in this PR..

because if you print a path (which is what those sprintfs are doing) it prints double slashes.

PR #267, opened 2020-03-31
Add a way to extract fields easily from JWTs in nsc

### LINKED ISSUE #264: Add a way to get keys from objects for scripting
To facilitate scripting we need a way to have `nsc` write the public key of an object (operator, account, user, activation token). 

For example:
```bash
$ nsc print key ./nsc/accounts/nats/KO/accounts/SYS/SYS.jwt
ABQKAY62555JHCLGRBRQSEAVXI5VSTMZIMLYXDYI6WCYZWD4ISEECZ2Y
```

This would allow scripts to direct output to env vars, passed to helm charts, etc, avoiding manual copy-and-paste in setting up system accounts, etc.

aricart:
Fix #264 - added `--json` flag to describe, which will print a json of the body in the JWT. This allows tooling such as jq to easily extract any field in the body of the JWT:
`nsc describe operator --json | jq .iss`
`nsc describe account --json | jq .sub`


cmd/describeaccount_test.go
```diff
@@ -166,3 +167,20 @@ func TestDescribeAccount_Latency(t *testing.T) {
...
+	out, _, err := ExecuteCmd(rootCmd, "describe", "account", "--json")
+	// reset the global
+	Json = false
+	require.NoError(t, err)
+	m := make(map[string]interface{})
+	err = json.Unmarshal([]byte(out), &m)
+	require.NoError(t, err)
+	ac, err := ts.Store.ReadAccountClaim("A")
+	require.NoError(t, err)
+	require.Equal(t, ac.Subject, m["sub"])
```
kozlovic, line 185: In other tests you would have the equivalent of `require.NotNil(t, ac)` here..

cmd/describe.go
```diff
@@ -62,6 +69,28 @@ var describeCmd = &cobra.Command{
 
 func init() {
 	GetRootCmd().AddCommand(describeCmd)
+	describeCmd.PersistentFlags().BoolVarP(&Json, "json", "J", false, "display JWT body as JSON")
 	describeCmd.PersistentFlags().BoolVarP(&WideFlag, "long-ids", "W", false, "display account ids on imports")
 	describeCmd.PersistentFlags().BoolVarP(&Raw, "raw", "R", false, "output the raw JWT (exclusive of long-ids)")
 }
+
+func bodyAsJson(data []byte) ([]byte, error) {
+	chunks := bytes.Split(data, []byte{'.'})
```
kozlovic, line 78: Since I don't know the format of `data` here, I can't really comment on the validity of this parsing.

cmd/describeaccount_test.go
aricart: I did it because the IDE was complaining that the value could be null - but it isn't. - the API returns null/empty values on error

cmd/describe.go
aricart: this is just JWT parsing - there are three JSON objects separated by '.'.

PR #269, opened 2020-04-02
added jsonpath support to describe to support tools

aricart:
added `--field` option, which loads the body of the jwt as JSON and then performs a JSON path evaluation.

The argument should be the field path separated by dots: `a.b.c[12].d` Note that you can reference indexed values etc.

### DIFF
--- a/cmd/jsonpath_test.go
+++ b/cmd/jsonpath_test.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2020 The NATS Authors
+ * Copyright 2020 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


cmd/jsonpath_test.go
```diff
@@ -0,0 +1,123 @@
...
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "unable to extract")
+}
+
+func TestSimple(t *testing.T) {
+	d := `{
+		"a":[1,"hello",true], 
+		"b": "hello",
+        "c": {"key": "one","value": "two"}
+}`
```
kozlovic, line 94: Could fix the tabulation for `d` content.

kozlovic:
@aricart If you have "global" flags that are used in tests, you should make sure that you run the test packages serially with the `-p=1` flag.

PR #271, opened 2020-05-06
FIX #263 Added ability to specify --account-jwt-server-url flag when pushing

### LINKED ISSUE #263: Add flag to set the account server url for `nsc push`
This would avoid having to hardcode an ip into the jwt when pushing or modifying /etc/hosts to the public ip in case the machine has not been deployed yet.

Instead of

```sh
nsc edit operator --account-jwt-server-url http://127.0.0.1:9090/jwt/v1/
nsc edit operator --account-jwt-server-url http://external-nats-account-server:9090/jwt/v1/ 
nsc push -A
```

Do for example:

```
nsc push -A --account-jwt-server-url http://127.0.0.1:9090/jwt/v1/
nsc push -A --account-jwt-server-url http://external-nats-account-server:9090/jwt/v1/ 
```

aricart:


cmd/push_test.go
```diff
@@ -16,6 +16,7 @@
 package cmd
 
 import (
+	"github.com/nats-io/jwt"
```
kozlovic, line 19: move down to other "github.com/..." import

PR #276, opened 2020-05-22
Add env variables to prevent certain user behaviours

### LINKED ISSUE #240: Flag to disable auto update and metadata
Implement option to be able to disable some of the metadata (nsc.json) file, git ignores and auto update that could help when scripting `nsc`.

aricart:
FIX #240 - Added env variables to prevent the creation of `.gitignore`, performing self-update checks, or writing ~/.nsc.json.

- `$NSC_CWD_ONLY` - when set the context is based on the location of the CWD (operator/account).
- `$NSC_NO_SELF_UPDATE` - when set will skip any self-update checks
- `$NSC_NO_GIT_IGNORE` - when set will not write .gitignore files in the directory pointed to by - `$NKEYS_PATH`. If the nkeys are in a directory under revision control, it means that the secrets will be checked in. Exercise caution.

### DIFF
--- a/cmd/options_test.go
+++ b/cmd/options_test.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2020 The NATS Authors
+ * Copyright 2020 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


cmd/options_test.go
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2018-2020 The NATS Authors
```
kozlovic, line 2: Date..

PR #278, opened 2020-05-29
adding windows to travis - fixing goveralls

aricart:
- Added windows build on travis
- Fixed missing goveralls

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,8 @@ script:
 
 deploy:
 - provider: script
-  script: $HOME/gopath/bin/goveralls -coverprofile=v2/coverage.out -service travis-ci
+  script: $HOME/gopath/bin/goveralls -coverprofile=coverage.out -service travis-ci
   on:
-    tags: true
     condition: $TRAVIS_OS_NAME = linux && $TRAVIS_GO_VERSION =~ ^1.14.
 
 - provider: script
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
 
 language: go
 go:
-- 1.12.x
+- 1.14.x
 
 git:
   depth: false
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -29,8 +29,13 @@ after_success:
 
 deploy:
 - provider: script
-  skip_cleanup: true
+  cleanup: false
```
kozlovic, line 32: May work, but I usually set `cleanup: true` which is an equivalent of `git stash`, basically cleaning the env. If you don't do that, the go releaser may fail to publish the release saying that git state is dirty.

.travis.yml
```diff
@@ -29,8 +29,13 @@ after_success:
...
   script: curl -sL http://git.io/goreleaser | bash
   on:
     tags: true
     condition: $TRAVIS_OS_NAME = linux
+- provider: script
+  script: $HOME/gopath/bin/goveralls -coverprofile=v2/coverage.out -service travis-ci
+  on:
+    tags: true
+    condition: $TRAVIS_OS_NAME = linux && $TRAVIS_GO_VERSION =~ ^1.12.
```
kozlovic, line 41: Newline maybe? Also, should you update the top and here to a more recent version of Go?

.travis.yml
aricart: fixed

.travis.yml
aricart: fixed

.travis.yml
```diff
@@ -16,22 +16,22 @@ install:
...
 script:
     go test -coverpkg=./... -coverprofile=./coverage.out ./...
 
-after_success:
-- git reset --hard
-
 deploy:
 - provider: script
-  skip_cleanup: true
+  script: $HOME/gopath/bin/goveralls -coverprofile=v2/coverage.out -service travis-ci
```
kozlovic, line 28: why `v2/coverage.out`?

.travis.yml
```diff
@@ -16,22 +16,22 @@ install:
...
+  on:
+    tags: true
```
kozlovic, line 30: So you would publish coverage only for tag pushes?

.travis.yml
aricart: c/p err

.travis.yml
aricart: changed to always

PR #283, opened 2020-06-17
Exposed account flag in delete user command

angiglesias:
Account Name flag (`-a`,`--account`)  of AccountContextParams wasn't added to cobra command flags


aricart:
@angiglesias yes I wasn't exposing it yet :)

aricart:
@angiglesias need a test where we are in a different account in the context, and the user is deleted there.

aricart:
Merged as #284. Thanks for the contribution.

aricart:
@angiglesias thank you!

PR #291, opened 2020-09-08
seed prints when invoking `nsc generate nkey --xxxx --store

### LINKED ISSUE #290: nsc generate nkey shoudn't print secret if --store is specified

aricart:
[fix] removed printing of seed key when storing a generated key. If the `--store` is not specified, the seed is printed.


aricart:
@philpennock if the --store option is provided it won't print anymore - you don't need to review the code, simply added you to ensure that you are aware I changed this.

matthiashanel (approved):
LGTM

PR #296, opened 2020-09-10
`generate profile` command

### LINKED ISSUE #287: "nsc generate profile"
Given an URL of the form:

`nsc://operator/account/user` resolve a JSON that looks like:

```json
{
  "user_creds": "<filepath>",
  "operator" : {
     "service": "hostport"
   }
}
```

The URL can provide arguments:
- `keystore=<dir>` that specifies the location of the keystore
- `store=<dir>` that specifies a directory that contains the named operator
- `[user|account|operator]Key` - includes the public key for user, account, operator,  If no prefix (user/account/operator is provided, it targets the last object in the configuration path)
- `[user|account|operator]Seed=<optional public key>` - include the seed for user, account, operator, if an argument is provided, the seed for the specified public key is provided - this allows targeting a signing key. If no prefix (user/account/operator is provided, it targets the last object in the configuration path)
- `[user|account|operator]Name` - includes the friendly name for the for user, account, operator,  If no prefix (user/account/operator is provided, it targets the last object in the configuration path)

`nsc://operator/account/user?seed`:
```json
{
  "operator": { "service": "host:port" },
  "user": {
    "seed": "SUXXXXXXXXX"
  }
}
```

NSC will offer an interface where it can request it as API, but will also allow you to do: `nsc generate profile <URL>` and spew the JSON out which can be redirected using `--out-file`.

aricart:
Added an internal `generate profile` mechanism to generate a JSON with interesting bits for tooling
FIX #287

### DIFF
--- a/cmd/generateprofile.go
+++ b/cmd/generateprofile.go
@@ -452,11 +452,11 @@ func (p *ProfileCmdParams) getKeys(claim jwt.Claims) []string {
 	if claim != nil {
 		keys = append(keys, claim.Claims().Subject)
 		var payload = claim.Payload()
-		oc, ok := payload.(jwt.Operator)
+		oc, ok := payload.(*jwt.Operator)
 		if ok {
 			keys = append(keys, oc.SigningKeys...)
 		}
-		ac, ok := payload.(jwt.Account)
+		ac, ok := payload.(*jwt.Account)
 		if ok {
 			keys = append(keys, ac.SigningKeys...)
 		}
--- a/cmd/generateprofile.go
+++ b/cmd/generateprofile.go
@@ -185,6 +185,24 @@ type NscURL struct {
 	qs       string
 }
 
+type stringSet struct {
+	set map[string]string
+}
+
+func newStringSet() *stringSet {
+	var s stringSet
+	s.set = make(map[string]string)
+	return &s
+}
+
+func (u *stringSet) add(s string) {
+	u.set[strings.ToLower(s)] = s
+}
+
+func (u *stringSet) contains(s string) bool {
+	return u.set[strings.ToLower(s)] != ""
+}
+
 func (u *NscURL) getOperator() (string, error) {
 	return url.QueryUnescape(u.operator)
 }
@@ -261,27 +279,17 @@ func (p *ProfileCmdParams) PostInteractive(_ ActionCtx) error {
 	return nil
 }
 
-func (p *ProfileCmdParams) hasName(name string, names []string) bool {
-	nn := strings.ToLower(name)
-	for _, n := range names {
-		if strings.ToLower(n) == nn {
-			return true
-		}
-	}
-	return false
-}
-
-func (p *ProfileCmdParams) loadNames(c jwt.Claims) []string {
-	var names []string
+func (p *ProfileCmdParams) loadNames(c jwt.Claims) *stringSet {
+	names := newStringSet()
 	cd := c.Claims()
-	names = append(names, cd.Name)
-	names = append(names, cd.Subject)
+	names.add(cd.Name)
+	names.add(cd.Subject)
 
 	conf := GetConfig()
 	payload := c.Payload()
 	_, ok := payload.(jwt.Operator)
 	if ok && conf.Operator != "" {
-		names = append(names, conf.Operator)
+		names.add(conf.Operator)
 	}
 	return names
 }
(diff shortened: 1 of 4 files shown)


aricart:
@ripienaar still in wip, needs more tests, is there anything else you would like in this?

cmd/resolve.go
```diff
@@ -0,0 +1,596 @@
...
+
+type Details struct {
+	Service []string `json:"service,omitempty"`
+	Name    string   `json:"name,omitempty"`
+	Seed    string   `json:"seed,omitempty"`
+	Key     string   `json:"id,omitempty"`
+}
+
+type ResolveResults struct {
+	UserCreds string   `json:"userCreds,omitempty"`
```
ripienaar, line 110: `user_creds` surely? :)

aricart: lol - that snake_casing_bad....

ripienaar: Just cos JS people cant make up their mind doesnt make it bad :P

aricart: lolz

aricart:
@ripienaar @matthiashanel renamed the command `profile` and hung it from `generate` - this seems more in line with it's intended use

ripienaar:
LGTM :)

cmd/resolve.go
```diff
@@ -0,0 +1,672 @@
...
+Other options (as query string arguments):
+keystore=<dir> that specifies the location of the keystore
+
+store=<dir> that specifies a directory that contains the named operator
+
+[user|account|operator]Key - includes the public key for user, account, 
+operator, If no prefix (user/account/operator is provided, it targets 
+the last object in the configuration path)
+
+[user|account|operator]Seed=<optional public key> - include the seed for
```
matthiashanel: list an example with the public key? I also can't find a unit test for this

cmd/resolve.go
```diff
@@ -0,0 +1,672 @@
...
+
+func (u *NscURL) getAccount() (string, error) {
+	return url.QueryUnescape(u.account)
+}
+
+func (u *NscURL) getUser() (string, error) {
+	return url.QueryUnescape(u.user)
+}
+
+func (u *NscURL) query() (map[Arg]string, error) {
```
matthiashanel: This looks like functionality that can be found in "net/url"
https://golang.org/pkg/net/url/#ParseQuery

cmd/resolve.go
```diff
@@ -0,0 +1,672 @@
...
+				return nil, err
+			}
+			v = s
+		}
+		m[Arg(k)] = v
+	}
+	return m, nil
+}
+
+func ParseNscURL(u string) (*NscURL, error) {
```
matthiashanel: I assume this id done manually as we don't use a host and so url parsing may be tricky?

cmd/resolve.go
```diff
@@ -0,0 +1,672 @@
...
+	names = append(names, cd.Subject)
+
+	conf := GetConfig()
+	payload := c.Payload()
+	_, ok := payload.(jwt.Operator)
+	if ok && conf.Operator != "" {
+		names = append(names, conf.Operator)
+	}
+	return names
+}
```
matthiashanel: map[string]struct{}?

cmd/resolve.go
aricart: done

cmd/resolve.go
aricart: Parse query is broken for arguments that don't have a value:
func (v Values) Get(key string) string {
	if v == nil {
		return ""
	}
	vs := v[key]
	if len(vs) == 0 {
		return ""
	}
	return vs[0]
}

cmd/resolve.go
aricart: correct - I am allowing spaces or URL encoded.

cmd/resolve.go
aricart: done

cmd/generateprofile.go
```diff
@@ -0,0 +1,680 @@
...
+type NscURL struct {
+	operator string
+	account  string
+	user     string
+	qs       string
+}
+
+type stringSet struct {
+	set map[string]string
+}
```
matthiashanel, line 190: type stringSet map[string]struct{} would have worked as well. Presence is the only thing checked.

matthiashanel (approved):
LGTM

assuming `resolve nsc://operator?userKey` would raise an error.

PR #309, opened 2020-10-26
[feat] use ascii tables

aricart:
changed tables to ASCII this allows for simpler and more consistent presentation, also matches the output from jetstream tools.


matthiashanel (approved):
LGTM

PR #310, opened 2020-10-27
fixed status adding nil values that had interface.

aricart:
Fixed check of nil on interface return value which bypasses the nil check to fail further down.


matthiashanel (approved):
LGTM, will add mine after

PR #316, opened 2020-11-02
Allow to overwrite an operator by same name and subject using --force

matthiashanel:


aricart (approved):
LGTM

cmd/addoperator.go
```diff
@@ -216,9 +218,15 @@ func (p *AddOperatorParams) PostInteractive(ctx ActionCtx) error {
 func (p *AddOperatorParams) Validate(ctx ActionCtx) error {
 	var err error
 	if p.token != "" {
+		if p.sysAcc {
+			return fmt.Errorf("importing an operator is not compatable with system account generation")
```
stjepang, line 222:
```suggestion
			return fmt.Errorf("importing an operator is not compatible with system account generation")
```

matthiashanel: fixed

matthiashanel: dang, didn't notice that commit suggestion button

PR #321, opened 2020-11-13
display list "users", "accounts" & operators in csv and save into a file

amitamit281:
Added option

nsc list operators -C
nsc list operators -C -F operators.csv

nsc list accounts -C -F accounts.csv
nsc list accounts -C -F accounts.csv

nsc list users -C -F users.csv
nsc list users -C -F users.csv

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -16,6 +16,7 @@
 package cmd
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"sort"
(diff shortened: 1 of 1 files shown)


aricart:
This is useful, however, can we add the command to do this on `generate` - something like `generate inventory --all | --operator --account --users --services --streams --csv` this way we could add other features as we move along.

aricart:
Also would need some tests verifying that this feature is doing work correctly.

aricart:
closing due to innactivity

PR #322, opened 2020-11-16
[feat] select from multiple keys when signing user

matthiashanel:


aricart (approved):
LGTM

PR #325, opened 2020-11-25
[added] check to protect against jwtv2 usage and provide instructions

matthiashanel:

added check to protect against importing account/operator/user jwtv2
added check to protect against using a jwtv2 store

### DIFF
--- a/cmd/addoperator.go
+++ b/cmd/addoperator.go
@@ -155,6 +155,9 @@ func (p *AddOperatorParams) Load(ctx ActionCtx) error {
 		}
 		op, err := jwt.DecodeOperatorClaims(token)
 		if err != nil {
+			if err.Error() == JWTV2DecodeError {
+				return fmt.Errorf("%sfailed to decode: %v", JWTUpgradeBannerJWT(), err)
+			}
 			return fmt.Errorf("error importing operator jwt: %v", err)
 		}
 		p.token = token
@@ -330,6 +333,9 @@ func (p *AddOperatorParams) Run(_ ActionCtx) (store.Status, error) {
 		if err == nil {
 			ocNew, err := jwt.DecodeOperatorClaims(p.token)
 			if err != nil {
+				if err.Error() == JWTV2DecodeError {
+					r.AddError("%sfailed to decode: %v", JWTUpgradeBannerJWT(), err)
+				}
 				return nil, err
 			}
 			if oc.Subject != ocNew.Subject {
--- a/cmd/importaccount.go
+++ b/cmd/importaccount.go
@@ -127,7 +127,11 @@ func (p *ImportAccount) Run(ctx ActionCtx) (store.Status, error) {
 	theJWT := p.content
 	claim, err := jwt.DecodeAccountClaims(string(theJWT))
 	if err != nil {
-		r.AddError("failed to decode %#q: %v", p.file, err)
+		if err.Error() == JWTV2DecodeError {
+			r.AddError("%sfailed to decode %#q: %v", JWTUpgradeBannerJWT(), p.file, err)
+		} else {
+			r.AddError("failed to decode %#q: %v", p.file, err)
+		}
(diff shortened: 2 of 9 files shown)


cmd/root.go
```diff
@@ -43,6 +43,24 @@ var cfgFile string
...
+In order to use jwt v2 you need to upgrade ALL your nats-server to be at least 2.1.9.
+`, GetRootCmd().Version, GetRootCmd().Version)
+}
+
+func JWTUpgradeBannerStore() string {
+	return fmt.Sprintf(`This version of nsc (%s) is incompatible with the stored operator version jwt.
+If you wish to proceed, you need to upgrade nsc with: "nsc upgrade".
+Note that you’ll always be able to return to this version by typing "nsc upgrade --version %s".
+In order to use jwt v2 you need to upgrade ALL your nats-server to be at least 2.1.9.
+`, "", "")
```
aricart, line 61: why sprint when there are no args...

cmd/addoperator.go
```diff
@@ -330,6 +333,9 @@ func (p *AddOperatorParams) Run(_ ActionCtx) (store.Status, error) {
 		if err == nil {
 			ocNew, err := jwt.DecodeOperatorClaims(p.token)
 			if err != nil {
+				if err.Error() == JWTV2DecodeError {
+					r.AddError("%sfailed to decode: %v", JWTUpgradeBannerJWT(), err)
```
aricart, line 337: possibly - remove the new line from the format, since you are just adding a line, and add the \n here.

cmd/addoperator.go
```diff
@@ -155,6 +155,9 @@ func (p *AddOperatorParams) Load(ctx ActionCtx) error {
 		}
 		op, err := jwt.DecodeOperatorClaims(token)
 		if err != nil {
+			if err.Error() == JWTV2DecodeError {
+				return fmt.Errorf("%sfailed to decode: %v", JWTUpgradeBannerJWT(), err)
```
aricart, line 159: possibly - remove the new line from the format, since you are just adding a line, and add the \n here.

cmd/importaccount.go
```diff
@@ -127,7 +127,11 @@ func (p *ImportAccount) Run(ctx ActionCtx) (store.Status, error) {
 	theJWT := p.content
 	claim, err := jwt.DecodeAccountClaims(string(theJWT))
 	if err != nil {
-		r.AddError("failed to decode %#q: %v", p.file, err)
+		if err.Error() == JWTV2DecodeError {
+			r.AddError("%sfailed to decode %#q: %v", JWTUpgradeBannerJWT(), p.file, err)
```
aricart, line 131: same as above - just add a `\n` in the template here

cmd/importuser.go
```diff
@@ -101,7 +101,11 @@ func (p *ImportUser) Run(ctx ActionCtx) (store.Status, error) {
 	}
 	claim, err := jwt.DecodeUserClaims(theJWT)
 	if err != nil {
-		r.AddError("failed to decode %#q: %v", p.file, err)
+		if err.Error() == JWTV2DecodeError {
+			r.AddError("%sfailed to decode %#q: %v", JWTUpgradeBannerJWT(), p.file, err)
```
aricart, line 105: and here...

aricart (changes requested):
Take a look at action.go, the `RunAction` and `RunStoreLessAction` and`RunMaybeStorelessAction` run all of our actions, this means you an intercept the error returned by any of the actions here, and issue the upgrade messages in one spot. If we are not chaining errors (and simply re-writing) we can possibly put that as one of the errors and have this thing process the list.

aricart:
Let's update jwt v1 to have a more descriptive error so we can catch any possible load, and the explicit checks will be taken care by this.

aricart (approved):
LGTM - for now.

PR #327, opened 2020-12-01
Migrate

matthiashanel:
I will merge these commits. For now I do not want to loose my history.
Basically I enforce that the operator jwt is v2 and provide guardrails and an upgrade routine.
Converting other jwt is not necessary.
I also do not write down the store version anywhere. Instead I rely on the operator jwt version.


matthiashanel:
```
> nsc upgrade-jwt
? It is advisable to create a backup of the store.
Do you want to create a backup in the form of a zip file now? Yes
? zip file name: /Users/matthiashanel/repos/nsc/DEMO-jwtV1-upgrade-backup.zip
Backup file "/Users/matthiashanel/repos/nsc/DEMO-jwtV1-upgrade-backup.zip"
created.
This backup does NOT CONTAIN private nkeys!
If you need to restore this state:
	Delete the directory "/Users/matthiashanel/test/demo./DEMO"
	Extract the backup
	Move the created directory in place of the deleted one
	Downgrade the nsc using: "nsc update -version <previous release>"
In order to use jwt V2, YOU MUST UPGRADE ALL nats-server prior to usage!
If you are NOT READY to switch over to jwt V2, downgrade nsc using:
"nsc update -version <previous release>"
? DID YOU UPGRADE nats-server AND DO YOU WANT TO CONVERT ALL NON MANAGED
OPERATOR / ACCOUNTS / USER JWT TO V2?
Once converted you need to re-distribute the operator jwt wherever used.
This includes, but is not limited to:
    ALL nats-server, which need to be restarted (one by one is ok)
    ALL dependent nsc stores in managed mode  Yes
[ OK ] Upgrade operator jwt to v2:
       [ OK ] Converted Operator: DEMO%
> nsc upgrade-jwt
[ OK ] Upgrade operator jwt to v2:
       [ OK ] Operator DEMO is already upgraded to version 2. No change was applied%
>
```

cmd/genericclaimparams.go
```diff
@@ -129,9 +129,24 @@ func (sp *GenericClaimsParams) Run(ctx ActionCtx, claim jwt.Claims, r *store.Rep
...
+	case jwt.OperatorClaim:
+		tags = &claim.(*jwt.OperatorClaims).Tags
+	case jwt.ActivationClaim:
+		tags = &claim.(*jwt.ActivationClaims).Tags
+	case jwt.AccountClaim:
+		tags = &claim.(*jwt.AccountClaims).Tags
+	case jwt.UserClaim:
+		tags = &claim.(*jwt.UserClaims).Tags
+	default:
+		panic("unhandled claim type")
```
aricart, line 144: This is only when a programming error when someone using the control yet it is not one of the above... Not sure we should have panic but also think that it will be fine.

aricart (approved):
LGTM - wondering about the messages that use all caps - I don't think we should be doing that.

PR #333, opened 2020-12-09
allow nsc update to work when encountering jwt v1 operator

matthiashanel:


aricart (approved):
LGTM

PR #335, opened 2020-12-16
[Added] command to validate jwt in a file

matthiashanel:
Input file can be a jwt, list of jwt (separated by new line)
If the jwt contains a | only the suffix will be interpreted as jwt
(result of pack call)
Also updated the jwt library.


aricart (approved):
LGTM

PR #341, opened 2021-01-07
Updating jwt library and nats-server (due to jwt lib)

matthiashanel:
Temporarily disabling staticcheck (will re enable when To to LocalSubject migration got added)


aricart (approved):
LGTM except for the merging issues. If you will address that else where feel free to merge.

PR #347, opened 2021-01-19
[added] option to import account signed by different operator

matthiashanel:


cmd/importaccount.go
```diff
@@ -46,6 +46,7 @@ func createImportAccountCmd() *cobra.Command {
 	cmd.Flags().BoolVarP(&params.skip, "skip", "", false, "skip validation issues, they can be edited out prior to push")
 	cmd.Flags().BoolVarP(&params.overwrite, "overwrite", "", false, "overwrite existing account")
 	cmd.Flags().StringVarP(&params.file, "file", "j", "", "account jwt to import")
+	cmd.Flags().BoolVarP(&params.force, "force", "", false, "import account signed by different operator")
```
aricart, line 49: I think this is what the migrate command does.

aricart (approved):
LGTM

cmd/importaccount.go
matthiashanel: the force there does smth else. account id is never checked.

PR #348, opened 2021-01-19
Improve component diagram: better display of imports without exports

matthiashanel:
and minor bug fixes and option to disable notesn

### DIFF
--- a/cmd/generatediagram.go
+++ b/cmd/generatediagram.go
@@ -36,6 +36,7 @@ func init() {
 		Args:         MaxArgs(0),
 		SilenceUsage: true,
 	}
+	accDetail := false
 	comp := &cobra.Command{
 		Use:          "component",
 		Short:        "Generate a plantuml component diagram for this store",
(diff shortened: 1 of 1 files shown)


cmd/generatediagram.go
```diff
@@ -151,18 +156,24 @@ skinparam component {
...
+    backgroundColor<<not-found public stream>> Red
+    backgroundColor<<not-found private stream>> Red
+}
 `)
 	addValidationNote := func(id string, name string, vr *jwt.ValidationResults) {
 		if len(vr.Issues) == 0 {
 			return
 		}
-		if len(vr.Issues) == 1 && strings.HasPrefix(vr.Issues[0].Description, "no activation provided for import") {
+		if len(vr.Issues) == 1 && strings.HasPrefix(vr.Issues[0].Description, "the field to has been deprecated") {
```
aricart, line 170: should this type of stuff be in a `nsc validate` - which then runs validations all we want on the store including all accounts?

aricart: I am thinking that otherwise we are adding all sorts of external validations - better - maybe we should do this as `warnings` from the JWT library...

aricart (approved):
LGTM

cmd/generatediagram.go
matthiashanel: These are warnings, just in this one context I don't want to show that one as it blows up the diagram.
I hope that no one will ever have to deal with that warnings as they use jwtv2 correctly.

PR #351, opened 2021-01-25
Minor fixes to make js work when only js params are provided

matthiashanel:
change unit converstion from 1000 to 1024 and removed printf


aricart (approved):
LGTM

PR #352, opened 2021-01-26
[added] operator edit  option to require signing keys in operator jwt

matthiashanel:
If on edit, accounts or users violating this rule exist the operation
fails.
Also generating system account with a signing key


aricart (approved):
LGTM

aricart (approved):
LGTM

PR #359, opened 2021-02-13
Minor fixes

matthiashanel:
[chore] remove mentioning of import prefix
[Added] support for rfc3339 formatted time when revoking
Fix js error when modifying the system account


aricart (approved):
LGTM

PR #360, opened 2021-02-16
Temporarily changed operator url back to v2

matthiashanel:
This will disable this version of nsc in ngs


aricart (approved):
LGTM

PR #361, opened 2021-02-16
Avoid always setting local subject and adding a prefix in interactive

matthiashanel:
Also straighten out table header


aricart (approved):
LGTM

aricart (approved):
LGTM

PR #366, opened 2021-03-09
Fixed issue where require strict usage was overwritten and push did fail

matthiashanel:
push was missing issuer account


aricart (approved):
LGTM

PR #367, opened 2021-03-09
Fixed linter issue and account server url not always being set

matthiashanel:


aricart (approved):
LGTM

PR #374, opened 2021-03-18
update nsc doc

aricart:


aricart:
updating nsc generated doc
[ci skip]

PR #387, opened 2021-05-03
[added] option to remove account and diff which accounts exist where

matthiashanel:

most of the change is moving code into separate functions.


aricart (approved):
LGTM

aricart:
There seems to be an issue with the static checks.

PR #390, opened 2021-05-07
`nsc pull -A` overwrites v2 operator with a v1

### LINKED ISSUE #388: nsc pull will replace a v2 operator with a v1
if you `nsc pull -A`, it is possible to pull a v1 operator `/jwt/v1/operator` instead of `/jwt/v2/operator` when the account server supports both.

Likely if the operator is v2 and the pulled operator is v2, the pull should be rejected (we may need a flag to specify the operator URL to pull).

aricart:
made operator pulls check v2 before v1, this is just a temporary solution, as the correct thing is to change the account server to handle a query string parameter to determine if the client is looking for a v2 JWT.
FIX #388

### DIFF
--- a/cmd/addoperator_test.go
+++ b/cmd/addoperator_test.go
@@ -227,7 +227,7 @@ func Test_AddWellKnownOperator(t *testing.T) {
 
 	// create the managed operator store
 	_, opk, okp := CreateOperatorKey(t)
-	as, _ := RunTestAccountServerWithOperatorKP(t, okp, 2)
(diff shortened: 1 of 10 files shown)


aricart:
The way to unwedge a replaced operator is to `nsc add operator --force -u https://api.synadia.io/jwt/v2/operator`

aricart:
Updated the well known operator for synadia to use the jwt/v2/operator.

matthiashanel (approved):
LGTM

PR #393, opened 2021-05-14
expanded on the help message provided for upgrading to v2. In the case of NGS it can be simpler

### LINKED ISSUE #392: help message for upgrading to jwt2 is not helpful enough

aricart:
FIX #392

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2020 The NATS Authors
+ * Copyright 2018-2021 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -128,9 +128,23 @@ var rootCmd = &cobra.Command{
...
-						return fmt.Errorf(`this version of nsc only supports jwtV2. To upgrade the v1 store %#q - type "%s upgrade-jwt". `+
-							`Alternatively you can downgrade %s to a compatible version using: "%s update -version 0.5.0"\n`,
-							store.GetName(), os.Args[0], os.Args[0], os.Args[0])
+						return fmt.Errorf(`this version of nsc only supports jwtV2. 
+If you are using a managed service, check your provider for 
+instructions on how to update your project. In most cases 
+all you need to do is:
+"%s add operator --force -u <url provided by your service>"
+
+If your services is well known, such as Synadia's NGS:
```
davidkemper, line 137: Should be "If your services are well known"...

davidkemper (changes requested):
Also, I happen to have a directory `~/synadia`, so when I tried the `--force -u synadia` option, it failed because that directory existed. The internals should be re-examined.

cmd/root.go
matthiashanel: Maybe stick to singular
If your service is well known

PR #396, opened 2021-05-14
[added] support for account key lookup by name in pubkeyparams

matthiashanel:


aricart (approved):
LGTM - save for probs the formatting

Issue #403, opened 2021-05-31
invalid memory address or nil pointer dereference (cmd/push.go:150)


mullerch:
I get this error when trying to `nsc push -A`:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9132a7]

goroutine 1 [running]:
github.com/nats-io/nsc/cmd.getSystemAccountUser(0xb25928, 0xc00021f050, 0xc000268278, 0x3, 0x0, 0x0, 0xa5bd16, 0x8, 0xc000263b98, 0x3, ...)
        /Users/aricart/Dropbox/code/src/github.com/nats-io/nsc/cmd/push.go:150 +0x807
github.com/nats-io/nsc/cmd.(*PushCmdParams).Run(0xc000170630, 0xb25928, 0xc00021f050, 0x0, 0x0, 0x0, 0x0)
        /Users/aricart/Dropbox/code/src/github.com/nats-io/nsc/cmd/push.go:544 +0x745
github.com/nats-io/nsc/cmd.run(0xb25928, 0xc00021f050, 0xa22c20, 0xc000170630, 0xb25928, 0xc00021f050)
        /Users/aricart/Dropbox/code/src/github.com/nats-io/nsc/cmd/action.go:144 +0x186
github.com/nats-io/nsc/cmd.RunAction(0xc000212500, 0xc00022a120, 0x0, 0x1, 0xa22c20, 0xc000170630, 0x0, 0x0)
        /Users/aricart/Dropbox/code/src/github.com/nats-io/nsc/cmd/action.go:112 +0xa5
github.com/nats-io/nsc/cmd.createPushCmd.func1(0xc000212500, 0xc00022a120, 0x0, 0x1, 0x0, 0x0)
        /Users/aricart/Dropbox/code/src/github.com/nats-io/nsc/cmd/push.go:47 +0x5e
github.com/spf13/cobra.(*Command).execute(0xc000212500, 0xc00022a100, 0x1, 0x1, 0xc000212500, 0xc00022a100)
        /Users/aricart/Dropbox/code/src/github.com/nats-io/nsc/vendor/github.com/spf13/cobra/command.go:762 +0x472
github.com/spf13/cobra.(*Command).ExecuteC(0xe1b7c0, 0xc000098000, 0x0, 0x0)
[...]

PR #404, opened 2021-06-01
The keystore API returns nil for a keypair and nil error if the key is not found

aricart:
Some of these cases on nsc have no integration tests (they require an account server)


aricart:
FIX #403

matthiashanel (approved):
LGTM

PR #407, opened 2021-06-22
Account token position

aricart:
[FIX] removed misleading flag help for service as only supported on public imports

[FEAT] Added support for `--account-token-position`
[FEAT] Added support for `--advertise` (hidden) for public exports.

The `--account-token-position` allows public exports from an account to be constrained to
a subject that is clamped to the account. The NATS server will ensure that only an
importing account sporting a matching public key in the specified wildcard token can
access the service/stream. This removes the burden of generating a token for subjects
that follow such a pattern on system-wide services


matthiashanel (approved):
LGTM

PR #409, opened 2021-07-01
[fixed] export/import type mismatch in diagram generation

matthiashanel:
now shows as not connected


aricart (approved):
LGTM

PR #415, opened 2021-07-13
[adding] support for scoped signing key

matthiashanel:

I believe user still have to improve the user


aricart (approved):
LGTM

PR #435, opened 2021-11-15
dependency and dependabot alerts

aricart:
[UPDATE] jwt to current version
[UPDATE] dependabot alert on github.com/ulikunitz/xz, and nats-server


kozlovic (approved):
LGTM

PR #436, opened 2021-11-30
fixed scoped signing key re-signing issue

matthiashanel:
Because the user aleady had an issuer, you could never re-sign with a
different key.


aricart (approved):
LGTM

PR #438, opened 2021-12-08
Adding --nats-resolver-cache option to generate config

matthiashanel:
This will generate the nats resolver config for the cache resolver


aricart (approved):
LGTM

PR #440, opened 2021-12-11
added code to keep plantuml from rendering * as O

matthiashanel:
`*` is some sort of special relationship notation for plantuml

without change 
Screen Shot 2021-12-10 at 8 33 58 PM

with change
Screen Shot 2021-12-10 at 8 34 52 PM


aricart (approved):
LGTM

PR #448, opened 2022-01-23
Normalize cli tool locations

aricart:
Previous version of nsc stored configuration on `~/.nsc` and data in `~/.nsc/nats` and keys in the directory pointed to by `$NKEYS_PATH`. This change adopts XDG:

- configuration moves to `$XDG_CONFIG_HOME` (`~/.config/nsc`)
- data moves to `$XDG_DATA_HOME` (`~/.local/share/nsc`) - with JWTs going to `~/.local/share/nsc/store`s, and keys to `~/.local/share/nsc/keys`

In addition, 3 new flags are available to all commands:

`--config-dir`, `--data-dir`, `--keystore-dir` these control the location of the config file, the data directory and the location of the keystore.

All the above paths are pre-calculated - previously some were calculated at runtime by examining environment variables.


aricart:
The currently submitted pr is excluding the vendor files, so the build will likely fail. Lots of changes on the tests were necessary because the tests simply assigned some of the locations and expected path names to have specific values, which makes changes such as this very difficult.

matthiashanel (commented):
A few questions:
When I migrate two store directories, will there be conflicts if they corresponded to the same keys directory? Basically can the same key be migrated twice without conflict? What if in two store directories the same operator and account names where used?

aricart:
> A few questions: When I migrate two store directories, will there be conflicts if they corresponded to the same keys directory? Basically can the same key be migrated twice without conflict? What if in two store directories the same operator and account names where used?

I have not implemented what this means - but effectively all I wanted to do was warn about ~/.nsc/... and migrate that to the new location. Same strategy for ~/.nkeys, this means that other locations would be left untouched. The reason for this migration is because then we can intercept the move and update nats cli contexts (maybe).

The migrate command should only be used to move the default locations, full stop.

kozlovic (approved):
LGTM

matthiashanel (commented):
LGTM

PR #451, opened 2022-02-01
Windows PowerShell Installer

scottf:


derekcollison:
Do we have a way to test this on a windows machine? I have one and can give it a try if I know how to test and what I should look for afterwards to denote success.

scottf:
> Do we have a way to test this on a windows machine? I have one and can give it a try if I know how to test and what I should look for afterwards to denote success.
I developed it on windows. 

* You can test it by making sure NSC is not in the path on your windows machine, run the script with no parameters, seeing it output the `nsc --version` result and running it yourself from the current or a new PowerShell window. 
* The latest version will be installed in your `$home\.nats\bin` directory, you should see nsc.exe
* You can run the script a second time to make sure it does not overwrite the existing exe.
* You can run the script with a version string i.e. `2.4.0` as a parameter to see that the correct version was installed. Make sure there is no exe already in place.
* You can run it with a bogus version, it fails with an error.
* You can delete the exe and re-run it again and check that the user PATH variable does not have the location listed more than once.

install.ps1
```diff
@@ -0,0 +1,78 @@
...
+$Version = ""
+if ($args.Length -eq 1) {
+    $Version = $args.Get(0)
+}
+
+$NscInstall = $env:NSC_INSTALL
+$BinDir = if ($NscInstall) {
+    "$NscInstall\bin"
+} else {
+    "$Home\.nats\bin"
```
aricart, line 25: The install location should be `$HOME/.local/bin`

aricart:
For natscli we need to account for the version in the zip - had a discussion with @ripienaar and we don't want to duplicate the binaries for the sake of the scripts. This means that we need to determine the version of the latest tag as described here:

https://blog.markvincze.com/download-artifacts-from-a-latest-github-release-in-sh-and-powershell/

We should put the logic of searching for the artifact with a version name and without - because if we do this, the script will only have to change the repository URL and be reusable for anything we want.

install.ps1
```diff
@@ -0,0 +1,78 @@
...
+    New-Item $BinDir -ItemType Directory | Out-Null
+	Write-Host "NSC directory ($BinDir) created...`r`n"
+}
+
+# Download the archive with the exe from github
+## GitHub requires TLS 1.2
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
+## Uri is different when a version is requested
+if (!$Version) {
```
aricart, line 47: I know that for nsc we don't bake the version in on the zip - but for other things we do - we should make this generic, and attempt both of the file names. To determine the tag on latest take a look at 

https://blog.markvincze.com/download-artifacts-from-a-latest-github-release-in-sh-and-powershell/

aricart (changes requested):

install.ps1
scottf: 1. By windows convention, binaries belong in `C:\Program Files\<appname>\` (or sometimes `C:\Program Files\<appname>\bin\`) when being made available for all users. When being added only for the current user, binaries go under `$Home\AppData\Local\<appname>\`

2. `.local` means nothing to Windows users and is confusing. Again, the Windows equivalent is `$Home\AppData\Local\`

3. `$Home\.nats\bin\` is a compromise. As it is common to have settings and configuration for an app under `$Home\.<appname>\`, there will be some anchor for Windows users. Adding `bin` was what Deno did and it made sense to me to separate binaries and config. 

4. The CLI uses `$Home\.config\nats\`. The `.config` location is also a Unix concept and should not have been used for Windows. `$Home\.nats\` would have been better.

We don't force clients to be exactly identical, we use and expect language specific paradigms. The same thing should apply here and we should use OS specific paradigms.

tbeets:
LGTM

install.ps1
scottf: I'm pulling the information from Phil's [json file](https://get-nats.io/synadia-nats-platforms.json) that has the details we need.

PR #455, opened 2022-02-03
[FIX] handling of account names instead of public keys

### LINKED ISSUE #454: `add import -a A -t TargetName` fails to resolve `TargetName`
The `-t` flag can take a name or a public key - when used with a name, it will fail to resolve

aricart:
PubKeyParams had functionality that initialized when the command flags were preparing for a command invocation. If the context of the store was not properly set, it failed. The fix handles the initialization of the flags during the command execution by providing the context for the command. Also, functionality to handle usernames instead of keys would wasn't handled.

FIX #454

### DIFF
--- a/cmd/addimport.go
+++ b/cmd/addimport.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2021 The NATS Authors
+ * Copyright 2018-2022 The NATS Authors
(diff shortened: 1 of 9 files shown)


kozlovic (approved):
LGTM

Issue #429, opened 2021-09-16
when doing nsc add operator -U if the URL is bad the error message is non-sensical


aricart:
```
$ nsc add operator --name test-syn -u "https://jwt.ngs-api.stage.synadia-ops.com/jwt/v2/operator"
Error: error reading `https://jwt.ngs-api.stage.synadia-ops.com/jwt/v2/operator`: open /Users/x/https:/jwt.ngs-api.stage.synadia-ops.com/jwt/v2/operator: no such file or directory
```

Note the file prefix above... `/Users/x`

variadico:
Resolved via https://github.com/nats-io/nsc/pull/456

PR #456, opened 2022-02-05
show better error when adding operator w/ bad url

natw:
I think this should address https://github.com/nats-io/nsc/issues/429

Instead of throwing away the error from fetching the path as a url, hold on to it and if fetching as path also fails, return the first error.

### DIFF
--- a/cmd/addoperator.go
+++ b/cmd/addoperator.go
@@ -150,26 +150,17 @@ func (p *AddOperatorParams) Load(ctx ActionCtx) error {
 	}
 	if p.jwtPath != "" {
 		var err error
-		var urlLoadingErr error
 		var data []byte
 		loadedFromURL := false
 
-		// If we think the path is a url, try loading it as such.
-		// If that doesn't work, still try it as a path.
-		// If THAT doesn't work, and we have an error from the url reading step,
-		// that's the one we probably care about.
 		if IsURL(p.jwtPath) {
 			loadedFromURL = true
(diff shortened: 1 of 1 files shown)


cmd/addoperator.go
```diff
@@ -150,15 +150,24 @@ func (p *AddOperatorParams) Load(ctx ActionCtx) error {
...
 	if p.jwtPath != "" {
 		var err error
+		var urlLoadingErr error
 		var data []byte
 		loadedFromURL := false
+
+		// If we think the path is a url, try loading it as such.
+		// If that doesn't work, still try it as a path.
+		// If THAT doesn't work, and we have an error from the url reading step,
+		// that's the one we probably care about.
```
aricart, line 160: IMHO, this should just be this:

```
if IsURL(p.jwtPath) {
	loadedFromURL = true
	data, err = LoadFromURL(p.jwtPath)
} else {
	data, err = Read(p.jwtPath)
}
if err != nil {
	return fmt.Errorf("error reading %#q: %v", p.jwtPath, err)
}
```

aricart (changes requested):

cmd/addoperator.go
natw: I do like that better. I was just trying to preserve the current behavior where the path could be something that _looks_ like a URL, but is actually a valid file path. I'm not quite pedantic enough to believe that changing this would be breaking backwards-compatibility, but I sympathize with it.

I leave that call to you :)

aricart (approved):
LGTM

PR #462, opened 2022-02-23
Make releases in GitHub Actions on tag push

philpennock:
* Adjust goreleaser to make reproducible builds, so that others can verify that the binaries we make were built from the source we provide
* Keep all the default linker variables from goreleaser in the binary, even if we only pass the version to cobra
* Adjust the checksums filename to be explicit that it has SHA256 checksums in it, so that we can add other types in future if needed
* Sign the checksum files with two approaches, matching the new installer
* Document here the signing and keys and how it fits together
* Put `nsc` into `.gitignore` so that a `go build` won't leave git wanting to add the binary.

The public keys are kept here inside the new `release/` top-level directory; the private keys are already in GitHub.

I made the indentation style in .goreleaser.yml be consistent, and reordered a little; diff with whitespace changes suppressed to review.


.github/workflows/release.yaml
```diff
@@ -0,0 +1,69 @@
+name: NSC client tool release
+on:
+  push:
+    tags:
+      - 'v*'
```
aricart, line 5: so nsc release tags do not have a `v` on them - not sure if this is an issue

aricart (approved):
LGTM if the workflow doesn't require the release tag to have a `v` as part of the tag name.

main.go
```diff
@@ -19,8 +19,15 @@ import (
 	"github.com/nats-io/nsc/cmd"
 )
 
-// the cli injects the version
-var version = "0.0.0-dev"
+// The build system should inject this data.
+// We only currently use the version, because of cobra's model, but we stamp
+// the rest into the binary so that debuggers can see it.
+var (
```
aricart, line 25: linter is failing the tests based on this.

.github/workflows/release.yaml
philpennock: How would we tell apart "this tag is for a release" from others?
Any reason to not start using a v in future?

main.go
philpennock: Fixed, thanks.

.github/workflows/release.yaml
philpennock: For the record here: because it will break the self-updater logic.  I added comments to this effect in the update commit.

PR #463, opened 2022-02-24
Switch to GitHub Actions for CI build checks

philpennock:
* We drop the goreleaser step which was non-functional on the previous CI, because that's already covered in a separate workflow YAML file, on tag push.
* We carefully make sure we still work with modules disabled.
  + Marked up lots of places to make it easier to remove all the added logic to support this
* We build on Ubuntu, macOS and Windows, but only do both current and previous Go versions on Ubuntu: I don't think we would get value out of testing prior-Go on all platforms, as long as we're testing on one.
  (There are scenarios where it would expose an issue, but not likely enough to be worth slowing things down for them.
  + The powershell differences around ${env:FOO} require some care to deal with
  + There was an issue with `go test -coverprofile` on Windows, so I ended up just using a different test invocation step on Windows.

For now, we don't fail-fast because the multi-platform support is new enough that I want to expose issues on all OSes, and see where things do work.  Once we're happy with the stability of this, we should remove the `fail-fast: false` directive.


aricart (approved):
LGTM

PR #469, opened 2022-03-08
Add nsc load cmd

wallyqs:
Given an URL it will bootstrap the store and keys folder structure and create a default user for an account plus a configured nats context.

Well known operators:

```sh
nsc load --profile "nsc://synadia?secret=SAABONGWSU7EWYZV3VAFFSP45P7NG3G67OLVRMA56UGZ2B3IYOGXAIFHLQ&user=bar"
```

Custom operators:

```sh
nsc load --url "http://localhost:9090/jwt/v1/operator" \
     --seed SAABONGWSU7EWYZV3VAFFSP45P7NG3G67OLVRMA56UGZ2B3IYOGXAIFHLQ \
     --user hello
```

Example usage:

```sh
$ nsc load --user hello2 --url "http://localhost:9090/jwt/v1/operator" --seed SAABONGWSU7EWYZV3VAFFSP45P7NG3G67OLVRMA56UGZ2B3IYOGXAIFHLQ
[ OK ] push jwt to account server
[ OK ] pull jwt from account server
[ OK ] generated and stored user key "UAIXTULQFMPW75TCE6TV7WYFLWQPXURNSNXRTQCJROSBPTQNTRV4DUQ5"
[ OK ] generated and stored user credentials at "/Users/waldemarquevedo/repos/nats-dev/src/github.com/nats-io/nsc/user-env/nsc/nkeys/creds/KO/A/hello2.creds"
[ OK ] created nats context "nsc-KO-hello2-user"

$ nats context list
╭──────────────────────────────────────────╮
│              Known Contexts              │
├────────────────────────────┬─────────────┤
│ Name                       │ Description │
├────────────────────────────┼─────────────┤
│ nsc-KO-hello2-user         │             │
╰────────────────────────────┴─────────────╯

$ nats context select nsc-KO-hello2-user 
NATS Configuration Context "nsc-KO-hello2-user"

      Server URLs: nats://127.0.0.1:4222
      Credentials: ~/src/github.com/nats-io/nsc/user-env/nsc/nkeys/creds/KO/A/hello2.creds (OK)
       NSC Lookup: nsc://KO/A/hello2
             Path: ~/src/github.com/nats-io/nsc/user-env/nsc/config/nats/context/nsc-KO-hello2-user.json

$ nats pub foo bar
13:20:56 Published 3 bytes to "foo"
```

Signed-off-by: Waldemar Quevedo <wally@synadia.com>

### DIFF
--- a/cmd/load.go
+++ b/cmd/load.go
@@ -1,24 +1,22 @@
-/*
- * Copyright 2018-2022 The NATS Authors
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// Copyright 2022 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
 package cmd
 
 import (
-	"errors"
 	"fmt"
-	"reflect"
+	"net/url"
+	"os/exec"
 	"strings"
 
 	"github.com/nats-io/jwt/v2"
@@ -49,7 +47,10 @@ func createLoadCmd() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVarP(&params.Profile, "profile", "p", "", "profile url")
+	cmd.Flags().StringVarP(&params.resourceURL, "profile", "", "", "Profile URL to initialize NSC and NATS CLI env")
+	cmd.Flags().StringVarP(&params.accountServerURL, "url", "", "", "URL of the account server")
+	cmd.Flags().StringVarP(&params.accountSeed, "seed", "", "", "Seed of the account used to create users")
+	cmd.Flags().StringVarP(&params.username, "user", "", "default", "Default username")
 	return cmd
 }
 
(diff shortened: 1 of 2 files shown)


cmd/action.go
```diff
@@ -61,7 +61,7 @@ func NewActx(cmd *cobra.Command, args []string) (ActionCtx, error) {
 	s, err := GetStore()
 	// in the case of add operator, there might not be a store
 	if err == ErrNoOperator {
-		if cmd.Name() == "operator" && cmd.Parent().Name() == "add" {
+		if cmd.Name() == "load" || (cmd.Name() == "operator" && cmd.Parent().Name() == "add") {
```
aricart, line 64: let's not change this, and simply add the "load" command as a top-level command that is not exposed.

cmd/wellknownoperator.go
```diff
@@ -35,9 +35,15 @@ type KnownOperators []KnownOperator
 var wellKnownOperators KnownOperators
 
 func defaultWellKnownOperators() KnownOperators {
-	return KnownOperators{
+	list := []KnownOperator{
 		{Name: "synadia", URL: "https://www.synadia.com", AccountServerURL: "https://api.synadia.io/jwt/v2/synadia"},
 	}
+	if os.Getenv("NSC_DEV") != "" {
```
aricart, line 41: the wellknown operator can be defined in the environment instead.

aricart (changes requested):

cmd/action.go
wallyqs: ok taking a look

cmd/wellknownoperator.go
wallyqs: thanks,  confirmed locally by setting `NSC_LOCAL_OPERATOR=http://localhost:9090/jwt/v1/operator` and it worked 👍

cmd/action.go
wallyqs: change to `action` was not needed so took out this change

wallyqs:
@aricart applied fixes to create the environment when there is no operator, it should be possible to create env from scratch now.

cmd/load.go
```diff
@@ -0,0 +1,366 @@
...
+			// Note: Need to initialize the operator and environment
+			// before the RunAction command since that is dependent
+			// on an operator being available.
+			params.setupHomeDir()
+			params.addOperator()
+			err := RunAction(cmd, args, &params)
+			if err != nil {
+				switch err.Error() {
+				case "operator not found":
+					cmd.Printf("Unable to find operator %q\n", params.operatorName)
```
aricart, line 44: should the message instead tell them to select the correct store or define the operator name in the environment as `NSC_<OPERATOR>_OPERATOR http/s://host:port/jwt/v1/operator`....

aricart (changes requested):
The creation of the store outside of the command flow, seems weird to me - as in I don't think the other commands (any of them) do that - but the command is working. I think the addoperator command may be doing something very similar to this - at some point I remember some storeless action (maybe that would solve the problem and allow you to run the without bypassing - I can get it to work correctly on a clean environment, but saw some situations where the error was simply an exit with a 1.

cmd/load.go
wallyqs: sounds good to me

cmd/load.go
```diff
@@ -0,0 +1,366 @@
...
+	accountSeed      string
+	accountPublicKey string
+	account          *jwt.AccountClaims
+	accountKeyPair   nkeys.KeyPair
+	accountServerURL string
+	username         string
+	contextName      string
+}
+
+func (p *LoadParams) setupHomeDir() {
```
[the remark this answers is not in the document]
wallyqs: updated to return errors instead

cmd/load.go
wallyqs: Updated error message:

```
Unable to find operator "KOo"

If you have used this operator, please enter:`nsc env -s /path/to/storedir`
or define the operator endpoint environment:
`NSC_<OPERATOR_NAME>_OPERATOR=http/s://host:port/jwt/v1/operator`

Error: operator not found
```

aricart (approved):
LGTM

Issue #471, opened 2022-03-23
the released packages do not have version info in the name


mcosta74:
The assets available for the releases just follow the pattern

- `nsc-<platform>.zip`: for binaries
- `nsc-<version>.zip/tar.gz`: for sources

would be nice to have also the version in the binary distribution:

`nsc-v2.6.1-darwin-amd64.zip` instead of `nsc-darwin-amd64.zip`

**NOTE**: you use the proposed naming schema in the `nats-server` repository

PR #472, opened 2022-03-23
Add version info to the packages name

mcosta74:
This PR add version information in the packages' name

`nsc-linux-amd64.zip` -->  `nsc-2.6.1-linux-amd64.zip` 

Closes: #471

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -63,7 +63,7 @@ dist: build
 archives:
   - id: "nsc.zip"
     wrap_in_directory: false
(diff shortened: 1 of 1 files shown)


aricart:
There's much more that needs doing for this to work, the self-update is looking for the versions in a specific way, so the change of the tag may create an issue here or there.

aricart:
@wallyqs can you take a look if we can simply just change the tag and not break the self-update (up or down.

mcosta74:
not 100% sure but the package `go-github-selfupdate` should be ok to download the right asset also if we change the archive name.

wallyqs:
We just started supporting tagged versions for Go modules, so we can look into supporting this for next release.

Issue #482, opened 2022-04-12
on an reset nsc environment, `nsc invoke failed` when doing a load


aricart:
bash-3.2$ export nsc_O_operator=https://nas.nas-dev.synadia.io/jwt/v1/operator
bash-3.2$ nsc load --profile "nsc://O?secret=SAAPZ7XT3T4QPUSFYJQ5SIMKTAEK6AZ5MMGKQ7KADYBNMNJFNB2WJGCBXA"
Error: nsc invoke failed: exit status 1
Usage:
  nsc load [flags]

Flags:
  -h, --help             help for load
      --profile string   Profile URL to initialize NSC and NATS CLI env
      --seed string      Seed of the account used to create users
      --url string       URL of the account server
      --user string      Default username (default "default")

Global Flags:
      --config-dir string     nsc config directory
      --data-dir string       nsc data store directory
  -i, --interactive           ask questions for various settings
      --keystore-dir string   nsc keystore directory
  -K, --private-key string    Key used to sign. Can be specified as role (where applicable),
                              public key (private portion is retrieved)
                              or file path to a private key or private key

bash-3.2$

wallyqs:
Fixed via #483

PR #483, opened 2022-04-13
Fixes to nsc load profile

### LINKED ISSUE #478: nsc load --profile & env variable issues
I used a web app which provided me with an `nsc load --profile` command to invoke; the app gave the `synadia` operator name in the nsc URN, although it was actually for a synadia-test stack.

When trying to use that:
1. Unless `nsc env -o foo` is used to set an operator name, `Error: set an operator -- 'nsc env -o operatorName'`
2. After setting the operator to an existing operator, where the operator JWT is on disk and contains an `account_server_url` field, that is not used and I still have to set an environment variable anyway
3. There's no handling of a hyphen in an operator name and the env var needs to be all upper-cased, so I had to switch `synadia-test` to `synadia_test`, and adjust the URL
4. With `nsc://synadia_test?` in the profile URN and `$NSC_SYNADIA_TEST_OPERATOR` in environment, but the `nsc env -o` value unchanged, the `nsc load --profile` invocation fails with an unhelpful `Error: nsc invoke failed: exit status 1`
5. After using `nsc env -o synadia_test` and repeating, everything worked.

wallyqs:
Adopt using `natscontext` lib instead of shelling out to `nats` to generate the context setup.

Fixes #478 #482


aricart (approved):
LGTM

PR #490, opened 2022-04-20
Don't ask GitHub Actions to upload brew formula

philpennock:
Cross-repo action access is an "interesting" security proposition right now.  Instead, skip the upload but register the formula as a build artifact, where it can be found to make a manual fix to the homebrew tap repo.

It's the least-bad solution.

WARNING 1: this is an _untested_ fix.  It looks sane to me.
WARNING 2: I'm about to head out the door for a long weekend vacation.  Merge or not as you see fit, to balance risk of "needing a release" vs "debugging if things go wrong", but since things are broken in GHActions right now without this, I don't think this will make things _worse_.

:speak_no_evil: famous last words.


wallyqs (approved):
LGTM

aricart (approved):
LGTM

PR #496, opened 2022-04-29
[FIX] added a validation check on the user context params

### LINKED ISSUE #491: Better error handling when "user" flag is missing
Steps to reproduce:
1. Use `nsc` command to pub/sub without the `--user` flag
```
nsc pub --account server1-admin public.a hello
```
Command tries to find the user creds file with the name ".creds"
```
Error: stat /Users/deepak.sah/.nkeys/creds/server1-op/server1-admin/.creds: no such file or directory: `/Users/deepak.sah/.nkeys/creds/server1-op/server1-admin/.creds`
```
Suggestion: If the `user` flag is missing or the creds file name is empty, return the appropriate message instead of trying to open the file.

aricart:
Added a missing validation check for the user context params. 
```
go run main.go pub -a A foo bar
Error: a user is required
Usage:
  nsc pub [flags]
```

FIXES #491

### DIFF
--- a/cmd/pubtool.go
+++ b/cmd/pubtool.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2019 The NATS Authors
+ * Copyright 2018-2022 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


matthiashanel (approved):
Lgtm

PR #499, opened 2022-05-11
FIX importing of accounts that are decorated

### LINKED ISSUE #498: `nsc import account` doesn't handle decorated jwts
```
nsc describe account A --raw --output-file a.jwt
nsc delete account A --force
nsc import account --file a.jwt
[ERR ] failed to decode `b.jwt`: illegal base64 data at input byte 10
```

aricart:
[FIX] import account read JWTs directly rather than use  `Read` which performs path expansions.
[FIX] read JWTs were not checked for decorations, making decorated JWTs fail the loading

FIX #498

### DIFF
--- a/cmd/importaccount.go
+++ b/cmd/importaccount.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2020 The NATS Authors
+ * Copyright 2018-2022 The NATS Authors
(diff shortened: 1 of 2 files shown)


matthiashanel (approved):
LGTM

PR #506, opened 2022-08-16
Update nats-io dependencies

aricart:
Preparing for release updated the nats-io stack and also bumped go versions.


kozlovic (approved):
LGTM

PR #512, opened 2022-08-26
Fix creation of extra nkey with --sk generate

JulienVdG:
The command `nsc edit operator --sk generate` currently save 2 nkeys in the store:
 - A first one when doing the Validate,
 - a second one when doing the Run.

This prevent the validate to generate a nkey and save it.


aricart (approved):

aricart:
But of course you are correct - I simply tested what the tool was reporting, a proper test shows that additional keys are added to the keystore.

aricart:
@JulienVdG thanks for the contribution.

PR #517, opened 2022-10-12
[FEAT] stream limits and tiered limits support

aricart:


cmd/editaccount.go
```diff
@@ -164,6 +183,165 @@ func (p *EditAccountParams) PreInteractive(ctx ActionCtx) error {
...
+			src = p.claim.Limits.JetStreamTieredLimits["R3"]
+		}
+		break
+	default:
+		return src, fmt.Errorf("unknown tier limit %q", tier)
+	}
+	return src, nil
+}
+
+//func (p *EditAccountParams) getTierSettings(tier string) (jwt.JetStreamLimits, *JetStreamLimitParams, error) {
```
kozlovic, line 271: Remove this?

cmd/editaccount.go
```diff
@@ -135,6 +145,8 @@ type EditAccountParams struct {
 	data           NumberParams
 	signingKeys    SigningKeysParams
 	rmSigningKeys  []string
+	r1             JetStreamLimitParams
```
kozlovic, line 148: I am not sure if this is correct to list only those. I mean technically one can have tiers per replicas count. The server determines if there are tiered limits based on the stream configuration's Replicas count.

kozlovic (commented):
Not sure if we limit to R1 and R3. Maybe user should be able to specify the limits to which tier to apply them to? but not sure how to make that command line friendly.

kozlovic (approved):
LGTM

PR #518, opened 2022-11-10
fix: nsc generate profile: use operator from URL

philpennock:
The `nsc generate profile` command should be able to emit JSON for any operator specified as a parameter; this is needed so that the natscli tool can use `nsc://` URLs in `$NATS_CONTEXT` definitions to reference users from any operator in the local store root.

Requiring that the correct operator already be the current operator defeats the whole point of being able to use `nsc://` URLs in the first place.

The `nsc generate profile` should not mutate the current profile.

This reworks the loading logic, without storing changes to the nsc.json settings, so that any known operator will work.

This does mean that various nsc settings to change where the store root is, or use CWD, will still impact upon the resolution of the URLs, but that's mostly reasonable: in the absence of overrides, nats can still reference multiple operators.

Without this change, `nats context ls` with contexts defined using `nsc:` URLs will almost certainly fail.  With this change, everything works in a way which satisfies my mental model.

### DIFF
--- a/cmd/generateprofile.go
+++ b/cmd/generateprofile.go
@@ -300,19 +300,37 @@ func (p *ProfileCmdParams) loadNames(c jwt.Claims) *stringSet {
 func (p *ProfileCmdParams) checkLoadOperator(ctx ActionCtx) error {
 	p.conf = GetConfig()
 	var err error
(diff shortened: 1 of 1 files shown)


aricart (approved):
LGTM

PR #519, opened 2022-11-11
Fix all staticcheck complaints

philpennock:
Two instances of a real bug were fixed, where a type-cast assignment shadowed the underlaying variable for the cast name, but the else-branch used the name assuming it was getting the non-shadowed variable instead of an empty variable.

Most of the other complaints were `io/ioutil` functions being deprecated.  There's a strings.Title deprecation too, even though we're only using English fixed strings.

Use of a variable `os` in code which imports package `os` suddenly became shadowing when switching `ioutil.` to `os.`, so I renamed `os` and then `as` for symmetry.

staticcheck built at dominikh/go-tools@4229c3d955f81fd39462d5eea9bc30a4f04a09d0


aricart (approved):
LGTM

PR #527, opened 2022-12-06
[fix] Tiered limits validation from blocking legitimate updates

codegangsta:
The new tiered limits feature added in #517 causes a validation failure
when a tier is set and a non-limits based flag is passed to update an
account.

This was discovered when trying to add a signing key to a fresh NGS
account.


codegangsta:
@aricart we may want to cut a release with this fix as it currently prevents NGS customers from adding signing keys to their accounts. I think imports/exports are okay, since they go through a different command, but any other claims like name, description, infoUrl are also affected

codegangsta:
No problem. Thanks for the response @philpennock

PR #528, opened 2022-12-07
support for xkey in the authcallout configuration

aricart:
- [FEAT] added `--curve`, `--rm-curve` key options to `authcallout` command
- [FEAT] updated describer to display curve key
- [FEAT] updated key list to show curve keys associated with an account


cmd/describeaccount_test.go
```diff
@@ -226,13 +226,16 @@ func TestDescribeAccount_Callout(t *testing.T) {
 
 	_, uPK, _ := CreateUserKey(t)
 	_, aPK, _ := CreateAccountKey(t)
+	_, xPK, _ := CreateNkey(t, nkeys.PrefixByteCurve)
```
derekcollison, line 229: CreateCurveKey()

derekcollison (commented):
Generally looks good. I wonder if we should move curve -> x25519? Maybe curve too generic?

cmd/describeaccount_test.go
aricart: Now that the nkey api correctly dispatches to the right API this is safe.

derekcollison (approved):
LGTM

PR #531, opened 2022-12-14
Switch module to v2

philpennock:
This is all that's needed for a build to work with a v2.7.5 tag laid down.

We could also change the import paths within the code to specify v2 and drop the `replace ... => ./` line.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
-module github.com/nats-io/nsc
+module github.com/nats-io/nsc/v2
 
 go 1.19
+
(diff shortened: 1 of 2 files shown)


philpennock:
EITHER this PR OR #532 should be used.

wallyqs (approved):
LGTM

wallyqs:
Let's merge this one instead: https://github.com/nats-io/nsc/pull/532

wallyqs (changes requested):
Let's use this one instead => https://github.com/nats-io/nsc/pull/532

PR #532, opened 2022-12-14
Switch module to v2 properly

philpennock:
Rewrite all the import paths.  Mostly `gofmt` replacement directives, and then some light editing to restore import groups.

Update Go dependencies.

This replaces #531 and we should use either that or this.


wallyqs (approved):
LGTM!!

PR #533, opened 2022-12-15
bump goreleaser-pro to 1.13.1-pro

philpennock:
This is the version I was testing with

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -61,6 +61,10 @@ jobs:
           # We have seen a couple of bugs, so are locking this for now.
           # v1.8.3-pro fixes some issues in brew architecture image variants
           # which led to <https://github.com/nats-io/homebrew-nats-tools/issues/9>
+          #
+          # 2022-12-14: testing carefully for the 2.7.5 -> v2.7.6 transition,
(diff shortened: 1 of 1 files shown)


wallyqs (approved):

.github/workflows/release.yaml
```diff
@@ -61,7 +61,11 @@ jobs:
           # We have seen a couple of bugs, so are locking this for now.
           # v1.8.3-pro fixes some issues in brew architecture image variants
           # which led to <https://github.com/nats-io/homebrew-nats-tools/issues/9>
-          version: v1.8.3-pro
+          #
+          # 2022-12-14: testing carefully for the 2.7.5 -> v2.7.6 transition,
+          # (v prefix, and v2 in the go modules), we used "1.13.1-pro" for testing,
+          # so lets make sure that we build with the version we tested worked.
+          version: 1.13.1-pro
```
wallyqs, line 68: now the Github action does not require the `v`? `v1.8.3-pro` -> `1.13.1-pro`

philpennock: I think that was a mistake on my part.  I think also that we don't need the comment addition, so I've updated to resolve the conflict, but at this point I think we can just drop this instead.

PR #537, opened 2023-01-20
Set Default output to help

sethjback:
This PR changes the default output of `nsc` from printing the version to showing the help.


aricart (approved):
LGTM

PR #539, opened 2023-01-21
chore: fix example for `describe jwt`

jonaslagoni:
The current example is stated as:

```
$ nsc describe jwt
Error: file is required
Usage:
  nsc describe jwt [flags]

Examples:
nsc describe -f pathorurl
```

When the example should be `nsc describe jwt -f pathorurl` as the other command is non-existing


aricart (approved):
LGTM

aricart:
@jonaslagoni thank you!

PR #540, opened 2023-01-25
Add `select` Command

sethjback:
This PR adds the `select` verb to the `nsc` command. The goal was to more explicitly expose the functionality to switch operators and accounts and to make it as easy as possible.


aricart (approved):
LGTM - only a minor nit, in nsc we use the `require` library for test assertions, it would be good if we kept that uniformity for now.

PR #541, opened 2023-01-25
Fix: null pointer error in add user interactive mode

sethjback:
When an operator is configured to require a signing key (`StrictSigningKeyUsage`) but there are no signing keys configure on an account, using interactive mode to create a user results in a null pointer error: the resolved signing key slice is empty and interactive mode attempts to auto select index 0 which doesn't exist.

This PR will return a descriptive error to the user indicating that signing keys are required but no signing keys are available for use.


aricart (approved):
LGTM

Issue #428, opened 2021-08-24
Provide a utility to bundle credentials for transfer between machines


ColinSullivan1:
From an email chain with a user:

```text
"copying working credentials sanely between machines"
is a use-case which some folks are actively working on
improving right now...

Oh wow, this would be really helpful. This is currently really complicated and error-prone.
```

One option might be a way to tar/zip all or a subset of credentials and then untar/unzip onto the target machine in the correct location. A really nice to have would be encrypting and decrypting the tarball/zip.

aricart:
@ColinSullivan1 how are they currently handling credential generation?
`nsc generate -a A -u U -o /some/file/path` is the right workflow.

Bulk copying credentials in a tar ball means that the user wants all the credentials (so can possibly have a bigger blast radius than expected). For that use case, they can just grab the credentials from `${NKEYS_PATH}/creds/<operator>/<account>`

aricart:
As for transferring/encrypting workflow, there are better tools such as ansible, etc to provision a machine, while using some mechanism for storing. For encrypting, the most secure mechanism is for the client to provide a user public key, and have the account 'manager' generate an user JWT for that user, this way the secret never leaves it's owner.

ColinSullivan1:
I didn't share enough about the use case - sorry about that.  The higher level use case is transferring private credentials from machine to machine for use with `nsc`.  I think a utility to do this would be less error prone than copying directories or regenerating user credentials.  Could see a case for wanting to copy all credentials, specific operator, accounts, or even users.  In this case, the user didn't know to copy the operator jwt/creds resulting in `nsc` not working on the target machine.  We haven't documented a process to do this, and creating a utility to do this would help users do this and allow for directory structure / file changes in the future without impacting users.

The original question:
```
What is the correct procedure to share those credentials with my team, i.e. how to I import those into nsc on a different computer? ```

aricart:
Added `--include-jwt` to the keys export - this puts all the keys and their associated JWTs in a directory. The jwts are named after the identity - as this insures uniqueness - creating an environment with `nsc fix --in dir...` which would on-board all the assets into a keystore/etc creating a clean env with just these assets.

PR #551, opened 2023-02-24
[FEAT] export an entire environment (jwts and keys)

aricart:
added `--include-jwtws` option to make it easy to export not only the keys, but also the configurations in the current operator tree. Paired with `nsc fix` you can recreate/mode an nsc environment.


aricart:
FIXES #428

piotrpio (approved):
LGTM!

PR #552, opened 2023-02-25
[FEAT] added single flag to set all the data/keys/config to one containing dir

aricart:
added `--all-dirs` option to store config, data, keys under the same directory - this is useful for generating test configurations

### DIFF
--- a/cmd/env_test.go
+++ b/cmd/env_test.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2022 The NATS Authors
+ * Copyright 2018-2023 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


cmd/env_test.go
```diff
@@ -71,3 +72,17 @@ func TestEnv_FailsBadAccount(t *testing.T) {
 	require.Error(t, err)
 	require.Contains(t, err.Error(), "\"A\" not in accounts for operator \"O\"")
 }
+
+func TestAllDir(t *testing.T) {
+	p := MakeTempDir(t)
+	defer func() {
+		os.RemoveAll(p)
+	}()
```
piotrpio, line 80:
```suggestion
	defer os.RemoveAll(p)
```

cmd/root.go
```diff
@@ -33,6 +33,7 @@ import (
 var ConfigDirFlag string
 var KeysDirFlag string
 var DataDirFlag string
+var AllDirFlag string
```
piotrpio, line 36: I would suggest keeping all config flags under one `var` block for clarity:
```
var (
  ConfigDirFlag string
  KeysDirFlag string
  DataDirFlag string
  AllDirFlag string
  ...
)
```

cmd/root.go
```diff
@@ -256,6 +263,6 @@ func HoistRootFlags(cmd *cobra.Command) *cobra.Command {
 	cmd.PersistentFlags().StringVarP(&ConfigDirFlag, "config-dir", "", "", "nsc config directory")
 	cmd.PersistentFlags().StringVarP(&DataDirFlag, "data-dir", "", "", "nsc data store directory")
 	cmd.PersistentFlags().StringVarP(&KeysDirFlag, "keystore-dir", "", "", "nsc keystore directory")
-
+	cmd.PersistentFlags().StringVarP(&AllDirFlag, "all-dirs", "H", "", "sets --config-dir, --data-dir, and --keystore-dir to the same value")
```
piotrpio, line 266: Why `-H` as the shorthand?

aricart: "home" - but that means something so I didn't want to make the human flag that

piotrpio (approved):
LGTM

PR #554, opened 2023-02-25
prevent entity names with path separators

### LINKED ISSUE #398: Role names should be validated
I was looking up what, if any constraints there are with role (operator, account, or user) names and so I tried this.

```
❯ nsc add account foo/bar
[ OK ] generated and stored account key "ACCPV5QZDNXZK3SJSN5RSHO3TVLKPH5GBIXRMAZZU4FJOCYAWDQ3XV3G"
[ OK ] added account "foo/bar"
Error: "foo/bar" not in accounts for operator "stage" in `/Users/ruthb/.nsc/nats`
```

aricart:
[FIX] entity names using a path separator are parsed as a path, which then writes them to the wrong location in the store - this fix generates an error if for such names

FIXES #398

### DIFF
--- a/cmd/addaccount.go
+++ b/cmd/addaccount.go
@@ -180,6 +180,11 @@ func (p *AddAccountParams) Validate(ctx ActionCtx) error {
 		return fmt.Errorf("account name is required")
 	}
 
+	if strings.Contains(p.name, "/") || strings.Contains(p.name, "\\") {
+		ctx.CurrentCmd().SilenceUsage = false
+		return fmt.Errorf("name cannot contain '/' or '\\'")
+	}
+
 	if p.name == "*" {
 		p.name = GetRandomName(0)
 	}
--- a/cmd/addaccount.go
+++ b/cmd/addaccount.go
@@ -180,7 +180,7 @@ func (p *AddAccountParams) Validate(ctx ActionCtx) error {
 		return fmt.Errorf("account name is required")
 	}
 
-	if strings.Contains(p.name, "/") || strings.Contains(p.name, "\\") {
+	if strings.ContainsAny(p.name, "/\\") {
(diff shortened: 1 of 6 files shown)


cmd/addaccount.go
```diff
@@ -180,6 +180,11 @@ func (p *AddAccountParams) Validate(ctx ActionCtx) error {
 		return fmt.Errorf("account name is required")
 	}
 
+	if strings.Contains(p.name, "/") || strings.Contains(p.name, "\\") {
```
piotrpio, line 183: You can use `ContainsAny` here (same for other usages in this PR):
```suggestion
	if strings.ContainsAny(p.name, "/\\")

File: cmd/adduser.go [hunk 2]
```diff
@@ -213,6 +213,11 @@ func (p *AddUserParams) Validate(ctx ActionCtx) error {
 		return fmt.Errorf("user name is required")
 	}
 
+	if strings.Contains(p.userName, "/") || strings.Contains(p.userName, "\\") {
```
piotrpio, line 216:
```suggestion
	if strings.ContainsAny(p.userName, "/\\") {
```

cmd/addoperator.go
```diff
@@ -248,6 +249,10 @@ func (p *AddOperatorParams) Validate(ctx ActionCtx) error {
 		ctx.CurrentCmd().SilenceUsage = false
 		return fmt.Errorf("operator name is required")
 	}
+	if strings.Contains(p.name, "/") || strings.Contains(p.name, "\\") {
```
piotrpio, line 252:
```suggestion
	if strings.ContainsAny(p.userName, "/\\") {
```

piotrpio (approved):
LGTM!

Issue #521, opened 2022-11-14
Imports with the same name seem to get confused


dylanratcliffe:
## Situation

My app has “customers”, each of whom has their own account to keep them nice and separate. There is however a shared database service that customers can make use of. This works as follows:

**To add something to the database:** The customer sends the item to the subject `streams.database` (This is a private stream, exported by the customer, imported by the database)

The database runs as a separate account, mapping the subjects as follows:

The `streams.database` stream is imported as `ingest.{customerName}`

The `{customerName}` is used in the subjects so that the database knows which customer a query has come from.

## Problem

When adding a *second* import to the database account using interactive mode NSC claims:

```
Error: account already imports stream "streams.database" from ABJX476VE7NKC5IBBNEWAVAWNNXW3W7P3WBP22QYZ2EHAQK73O27CJGK
```

Which is not true.

## Replication

Creating accounts:

```
vscode ➜ /tmp $ nsc init
? enter a configuration directory ./
? Select an operator Create Operator
? name your operator, account and user example
[ OK ] created operator example
[ OK ] created system_account: name:SYS id:ADAUPTVMMVDOTX34LGHN6S4GQMDUYVAPOJ477MPWEODOOFVKFT5OZXFC
[ OK ] created system account user: name:sys id:UC6CFISTAB56PW7YQ22IDRUC7HFX4QVTJ3FOH763OB4W6RKRXDANTART
[ OK ] system account user creds file stored in `~/.local/share/nats/nsc/keys/creds/example/SYS/sys.creds`
[ OK ] created account example
[ OK ] created user "example"
[ OK ] project jwt files created in `/tmp`
[ OK ] user creds file stored in `~/.local/share/nats/nsc/keys/creds/example/example/example.creds`
> to run a local server using this configuration, enter:
>   nsc generate config --mem-resolver --config-file <path/server.conf>
> then start a nats-server using the generated config:
>   nats-server -c <path/server.conf>
all jobs succeeded

vscode ➜ /tmp $ nsc create account database
[ OK ] generated and stored account key "ADLKTPY4MPF7JPF3N7C4TWSOU24ZD5XFQ573LKSD5GVMWPIWFL2LAPTT"
[ OK ] added account "database"

vscode ➜ /tmp $ nsc create account customerA
[ OK ] generated and stored account key "AA4XIPKG4D7O33A3HPTWQC7UZMQQAVQIPBVVEY2PUOHK27OOMVTSQ7HE"
[ OK ] added account "customerA"

[...]

PR #555, opened 2023-02-25
[FIX] interactive imports when two accounts export same subject confuse the source account selection

aricart:
[FIX] imports with the same name from different accounts get confused when using interactive mode - the cli ui is comparing strings and returns the last one match rather than the one at the current selection. The fix simply adds the src account name which disambiguates the entry


aricart:
FIXES #521

piotrpio (approved):
LGTM!

PR #556, opened 2023-02-26
Set initial context of generate profile to the provided nsc url

### LINKED ISSUE #508: Have nsc URL operator, account, user take precedence over environment
I observed the behavior of trying to generate a profile, e.g. `nsc generate profile nsc://hub/APP/user` while my nsc environment was set to a different operator. The command failed in the [validation check](https://github.com/nats-io/nsc/blob/main/cmd/generateprofile.go#L295-L314).

Knowing there are several forms of the URL that can be specified, my expectation would be that if all components (operator, account, user) are specified in the URL, it would ignore whatever is set in `nsc env`. If any components are missing, then it would fallback to the env.

aricart:
[FIX] generate profile now sets the current context for the specified nsc:// url

[FEAT] added 'seeds', 'keys', 'names' to the options - this adds the specified values to all the entities affected by the specified nsc url

FIX #508

### DIFF
--- a/cmd/exportkeys.go
+++ b/cmd/exportkeys.go
@@ -138,6 +138,9 @@ func (p *ExportKeysParams) Run(ctx ActionCtx) (store.Status, error) {
 
 	sr := store.NewDetailedReport(true)
 	keys.KeyList, err = p.KeyCollectorParams.Run(ctx)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


piotrpio (approved):
LGTM!

PR #561, opened 2023-02-27
[FIX] added optional timeout for nats-resolver based operations

### LINKED ISSUE #504: add argument to `nsc push` command to set the timeout
Hi team, nsc is very good tool.
But in my environment occurred timed out with `nsc push` frequently. 
Therefore, it would be nice if the timeout could be set from an argument.

aricart:
FIXES #504

### DIFF
--- a/cmd/push.go
+++ b/cmd/push.go
@@ -380,7 +380,7 @@ func (p *PushCmdParams) getSelectedAccounts() ([]string, error) {
 	return []string{}, nil
 }
 
-func multiRequest(nc *nats.Conn, report *store.Report, operation string, subject string, reqData []byte, respHandler func(srv string, data interface{}), timeout int) int {
+func multiRequest(nc *nats.Conn, timeout int, report *store.Report, operation string, subject string, reqData []byte, respHandler func(srv string, data interface{})) int {
 	ib := nats.NewInbox()
 	sub, err := nc.SubscribeSync(ib)
 	if err != nil {
@@ -446,14 +446,14 @@ func obtainRequestKey(ctx ActionCtx, subPrune *store.Report) (nkeys.KeyPair, str
 	return okp, opPk, nil
 }
 
-func sendDeleteRequest(ctx ActionCtx, nc *nats.Conn, deleteList []string, respList int, subPrune *store.Report, timeout int) {
+func sendDeleteRequest(ctx ActionCtx, nc *nats.Conn, timeout int, report *store.Report, deleteList []string, expectedResponses int) {
 	if len(deleteList) == 0 {
-		subPrune.AddOK("nothing to prune")
+		report.AddOK("nothing to prune")
 		return
 	}
-	okp, opPk, err := obtainRequestKey(ctx, subPrune)
+	okp, opPk, err := obtainRequestKey(ctx, report)
 	if err != nil {
-		subPrune.AddError("Could not obtain Operator key to sign the delete request (err:%v)", err)
+		report.AddError("Could not obtain Operator key to sign the delete request (err:%v)", err)
 		return
 	}
 	defer okp.Wipe()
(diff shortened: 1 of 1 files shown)


cmd/push.go
```diff
@@ -298,6 +298,12 @@ func (p *PushCmdParams) PostInteractive(_ ActionCtx) error {
 }
 
 func (p *PushCmdParams) Validate(ctx ActionCtx) error {
+	if p.timeout < 1 {
```
piotrpio, line 301: Don't you think it would be better to simply return error here to not confuse users? Especially for timeouts > 60s.

cmd/push.go
```diff
@@ -374,7 +380,7 @@ func (p *PushCmdParams) getSelectedAccounts() ([]string, error) {
 	return []string{}, nil
 }
 
-func multiRequest(nc *nats.Conn, report *store.Report, operation string, subject string, reqData []byte, respHandler func(srv string, data interface{})) int {
+func multiRequest(nc *nats.Conn, report *store.Report, operation string, subject string, reqData []byte, respHandler func(srv string, data interface{}), timeout int) int {
```
piotrpio, line 383: I would put `timeout` before `respHandler` to keep handler as the last param.

cmd/push.go
```diff
@@ -57,6 +55,7 @@ push -P -A (all accounts)`,
 	cmd.Flags().StringVarP(&params.removeAcc, "account-removal", "R", "", "remove specific account. Only works with nats-resolver enabled nats-server. Mutually exclusive of prune/diff.")
 	cmd.Flags().StringVarP(&params.sysAcc, "system-account", "", "", "System account for use with nats-resolver enabled nats-server. (Default is system account specified by operator)")
 	cmd.Flags().StringVarP(&params.sysAccUser, "system-user", "", "", "System account user for use with nats-resolver enabled nats-server. (Default to temporarily generated user)")
+	cmd.Flags().IntVarP(&params.timeout, "timeout", "", 1, "timeout in seconds [1-60] to wait for responses from the server (only applicable to nats-resolver configurations, and applies per operation)")
```
piotrpio, line 58: Please add note about the default value in description

piotrpio (changes requested):

cmd/push.go
aricart:
```bash
$ nsc push -h
Push an account jwt to an Account JWT Server

Usage:
  nsc push (currentAccount)
push -a <accountName>
push -A (all accounts)
push -P
push -P -A (all accounts) [flags]

Examples:
push

Flags:
  -a, --account string                  account name
  -u, --account-jwt-server-url string   set account jwt server url for nsc sync (only http/https/nats urls supported if updating with nsc) If a nats url is provided 
  -R, --account-removal string          remove specific account. Only works with nats-resolver enabled nats-server. Mutually exclusive of prune/diff.
  -A, --all                             push all accounts under the current operator (exclusive of -a)
  -D, --diff                            diff accounts present in nsc env and nats-account-resolver. Mutually exclusive of account-removal/prune.
  -F, --force                           push regardless of validation issues
  -h, --help                            help for push
  -P, --prune                           prune all accounts not under the current operator. Only works with nats-resolver enabled nats-server. Mutually exclusive of account-removal/diff.
      --system-account string           System account for use with nats-resolver enabled nats-server. (Default is system account specified by operator)
      --system-user string              System account user for use with nats-resolver enabled nats-server. (Default to temporarily generated user)
      --timeout int                     timeout in seconds [1-60] to wait for responses from the server (only applicable to nats-resolver configurations, and applies per operation) (default 1)
```

cmd/push.go
aricart: I think clamping for this is not a problem - simply doing a check in case they gave a negative value etc.

cmd/push.go
aricart: since that is effectively something related to the connection/etc (we need scatter gather in the go client), I put it after the connection, also refactored some of the param names so that we can understand what is going on in there.

piotrpio (approved):
LGTM

PR #565, opened 2023-03-01
Get jwt library v1 out of the nsc binary

philpennock:
* Add a CI push test to fail the build if the binary pulls in jwt library v1
  + this should be extensible enough as a pattern to let us handle any other deny-list libraries we want to keep out
* Fix `cmd/pubkeyparams.go` which accidentally pulled in jwtv1 not jwtv2, to get the constant `jwt.All` (`"*"`)
* Change all test files which didn't explicitly pull in jwt/v1 _as_ `jwtv1` to bring in jwt/v2 instead
  + If the test says jwtv1 explicitly, it's assumed to be deliberate

As long as any tests pull in JWTv1, the go.mod file will continue to reference v1, so we can't just audit the module file for sanity; we instead build the binary and look at what it's actually using.

(cherry picked from commit e677cb447f683f52adcf56cc95bbcef8e4c92107)
Signed-off-by: Phil Pennock <pdp@synadia.com>

### DIFF
--- a/.github/workflows/pushes.yaml
+++ b/.github/workflows/pushes.yaml
@@ -125,4 +125,19 @@ jobs:
           COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         if: matrix.canonical
 
+      - name: Bad versions creep defense
+        id: go-module-versions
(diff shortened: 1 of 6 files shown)


wallyqs (approved):
Excellent 💯

PR #566, opened 2023-03-01
bump all golang.org/x dependencies

philpennock:
We know we need some of these golang.org/x dependencies updated for security issues, and we also know that they're almost always safe with respect to backwards incompatibilities, so just bump up all these libraries to be current.


philpennock:
The push just now fixed the commit message and did not change the object tree.

go.sum
```diff
@@ -121,6 +121,8 @@ golang.org/x/crypto v0.0.0-20201221181555-eec23a3978ad/go.mod h1:jdWPYTVW3xRLrWP
 golang.org/x/crypto v0.0.0-20210314154223-e6e6c4f2bb5b/go.mod h1:T9bdIzuCu7OtxOm1hfPfRQxPLYneinmdGuTeoZ9dtd4=
 golang.org/x/crypto v0.4.0 h1:UVQgzMY87xqpKNgb+kDsll2Igd33HszWHFLmpaRMq/8=
 golang.org/x/crypto v0.4.0/go.mod h1:3quD/ATkf6oY+rnes5c3ExXTbLc8mueNue5/DoinL80=
+golang.org/x/crypto v0.6.0 h1:qfktjS5LUO+fFKeJXZ+ikTRijMmljikvG68fpMMruSc=
```
wallyqs, line 124: go mod tidy needed maybe?

philpennock: good catch, thanks, done

wallyqs (approved):
LGTM 🚀

PR #567, opened 2023-03-01
unbreak release tooling, versions bump

philpennock:
Most critically: cosign now requires --yes to confirm that it's okay for signatures to be uploaded, even when ... not using OIDC flows.  Sigh.

While doing releng treading-water tasks:
 * all future tags are v-preficed if they're for release
 * bump goreleaser to current
   + adjust goreleaser invocation/config for deprecation warnings

Intent to backport: this needs to be applied to rel-2.7 too


wallyqs (approved):
LGTM 👍

PR #570, opened 2023-03-16
bump dependencies

philpennock:
* nkeys to v0.4.4
* jwt/v2 to v2.4.0
* nats.go to v1.24.0
* adapt to cleaner JWTv1 import paths, get old v1 module out of our go.sum
* ran `go get -u` so have picked up all upstream version updates


aricart (approved):
LGTM

PR #572, opened 2023-03-23
[FIX] importing of user jwt files regardless of whether they are decorated or not

### LINKED ISSUE #571: nsc import user does not expect a decorated jwt
When importing a system account using the procedure described [here](https://docs.nats.io/running-a-nats-service/nats_admin/security/jwt#import-operator-self-service) the final step fails.

```
root@clientk1:~# nsc import user --file system_user.jwt
[ERR ] failed to decode `system_user.jwt`: illegal base64 data at input byte 10
```

I was able to import the user by editing the jwt to remove the decoration line at the top and the corresponding line at the bottom of the file.  That is, remove the "-----BEGIN NATS ACCOUNT JWT-----".

```
root@clientk1:~# nsc import user --file system_user.jwt
[ OK ] user sys-non-op was successfully imported
```

Note that `nsc add operator` and `nsc import account` expect decorated JWTs.

Similar issues reported in #381 #499.  Apologies if this is fixed already.

nsc was installed using:

```
go install github.com/nats-io/nsc@latest
```

aricart:
[TEST] added test for importing accounts regardless of whether they are decorated or not

FIX #571

### DIFF
--- a/cmd/importaccount_test.go
+++ b/cmd/importaccount_test.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2022 The NATS Authors
+ * Copyright 2018-2023 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


bruth (approved):
LGTM

PR #590, opened 2023-05-03
[BREAKING] this change directly writes action reports to stdout

### LINKED ISSUE #580: `nsc generate nkey` outputs good response to stderr rather than stdout
`nsc generate nkey` outputs a good response to stderr, rather than stdout, which is unexpected.      

```bash
root@instance:/# nsc generate nkey -a --store 2> e
root@instance:/# nsc generate nkey -a --store
AALS5NG2EJCJGCBTCRRH4PJO2R2UFHVAAIPLZEI5OPHHGW5T67DFNJRY
account key stored /root/.local/share/nats/nsc/keys/keys/A/AL/AALS5NG2EJCJGCBTCRRH4PJO2R2UFHVAAIPLZEI5OPHHGW5T67DFNJRY.nk
```

In particular, I would expect the following ansible to work.  But, to make it work i had to change stdout to stderr.

```yaml
    - name: Generate account signing key
      ansible.builtin.shell: >-
        nsc generate nkey -a --store
        {{ client_config }}
      register: result

    - name: Get the client account signing key 
      ansible.builtin.set_fact: client_account_signing_key="{{result.stdout_lines[0]}}"
```

aricart:
Fixes #580

### DIFF
--- a/cmd/action.go
+++ b/cmd/action.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 The NATS Authors
+ * Copyright 2018-2023 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


aricart:
https://github.com/nats-io/nsc/pull/696 does the right thing

PR #591, opened 2023-05-03
Clarify system account edits which requires JetStream limits to be set to 0

### LINKED ISSUE #569: nsc edit is checking if JetStream is available
When you enter the following command:
`nsc edit account SYS -k OD.....2Y --tag A`
You will get the following error:
`Error: jetstream not available for system account`
NSC-1

### LINKED ISSUE #580: `nsc generate nkey` outputs good response to stderr rather than stdout

aricart:
[FIX] clarified check added to edits of the system account which prevented editing due to JetStream default values in the flag - added a flag that sets the values for the jetstream setting (--js-disable) which sets the values to 0

Fixes #569

### DIFF
--- a/cmd/editaccount.go
+++ b/cmd/editaccount.go
@@ -117,7 +117,7 @@ func createEditAccount() *cobra.Command {
 	cmd.Flags().VarP(&params.MaxAckPending, "js-max-ack-pending", "", "JetStream: set number of maximum acks that can be pending for a consumer in the account")
 
 	cmd.Flags().StringVarP(&params.AccountContextParams.Name, "name", "n", "", "account to edit")
-	cmd.Flags().BoolVarP(&params.disableJetStream, "js-disable", "", false, "JetStream:  sets options --js-streams, --js-consumer, --js-mem-storage --js-max-mem-stream, --js-disk-storage, --js-max-disk-stream to 0")
+	cmd.Flags().BoolVarP(&params.disableJetStream, "js-disable", "", false, "disables all JetStream limits in the account by deleting any limits")
 
 	params.signingKeys.BindFlags("sk", "", nkeys.PrefixByteAccount, cmd)
 	params.TimeParams.BindFlags(cmd)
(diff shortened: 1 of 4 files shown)


bruth:
Is there any way to detect the account being edited is the system account for the selected/implicit operator?

aricart:
> Is there any way to detect the account being edited is the system account for the selected/implicit operator?

Fixed - now if the --js-xxx flags are not specified, the edit to the account is performed. Otherwise the offending flag(s) are rejected.

bruth (approved):
LGTM!

cmd/editaccount.go
```diff
@@ -236,6 +243,21 @@ func (p *EditAccountParams) getTierLimits(tier int) (*jwt.JetStreamLimits, error
 	return src, nil
 }
 
+func (p *EditAccountParams) doDisableJetStream(r *store.Report) error {
+	if !p.disableJetStream {
+		return nil
+	}
+	p.claim.Limits.JetStreamLimits = jwt.JetStreamLimits{}
+	r.AddOK("deleted global limit")
```
bruth, line 251: "delete" may be misinterpreted, I suggest "cleared global limits" and "cleared tiered limit"

cmd/editaccount.go
```diff
@@ -236,6 +243,21 @@ func (p *EditAccountParams) getTierLimits(tier int) (*jwt.JetStreamLimits, error
...
+	if p.claim.Limits.JetStreamTieredLimits != nil {
```
bruth, line 252: The nil check is not needed for iteration.

bruth (approved):
Two minor suggestions, but otherwise LGTM

cmd/editaccount.go
aricart: fixed

cmd/editaccount.go
aricart: in other operations and tiers it is also labeled as "deleted" - so keeping it that way for now.

PR #593, opened 2023-05-07
YAML Enginering: quote go-version string

philpennock:
Try to ensure that when we switch to Go 1.20 we don't forget to quote the string to protect against the version being parsed as `1.2`.


wallyqs (approved):

aricart (approved):
Seems crazy that quoting is required - but I guess that is what Semver solved :)

Issue #594, opened 2023-07-06
Generating of description for entities in combination of `json` and `write to file` is broken


Drakorgaur:
# Problem definition
In an attempt to create a description of an operator, account, or user using the `nsc` tool with the output in JSON format and writing the output to a file via the `--output-file` flag, the tool returns an error.

## Steps to reproduce:
- [ ] set up nsc(set env vars, prepare space, etc)
- [ ] create operator(account, user optional)
- [ ] run `nsc describe operator -J --output-file $(pwd)/test.json`

### Expected behaviour:
Stdout:
```bash
[ OK ] wrote operator description to `~/<omitted>/test.json`
```
### Real behaviour:
Stdout:
```bash
Error: expected 3 chunks
```

## Problem explaination
When  `cmd.Json` is set, Load method sets `raw` attribute as JSON-bytes
```go
func (p *DescribeUserParams) Load(ctx ActionCtx) error {
  . . .
  p.raw, err = bodyAsJson(p.raw)
  . . . 
}
```
after this in Run method tries to decode JWT:
```go
func (p *DescribeUserParams) Run(ctx ActionCtx) (store.Status, error) {
  . . . 
  p.raw, err = jwt.DecorateJWT(string(p.raw))
  . . . 
}
```
which expects bytes be separated by dot, but it's already decoded to JSON.

PR #595, opened 2023-07-06
fix: Generating of description for entities in combination of json and write to file.

Drakorgaur:
see issue https://github.com/nats-io/nsc/issues/594


aricart:
@Drakorgaur great find thanks

aricart (changes requested):
I'll add a test

aricart (approved):

Drakorgaur:
> The fix is super simple, can you add a test for it, if not I can add it for you

yes, no problem.
Can it wait for tommorow then?

aricart:
I'll do them! - it will be much less painful on you.

Drakorgaur:
> I'll do them! - it will be much less painful on you.

thanks! 😄

PR #611, opened 2023-09-23
[FEAT] expirations command

aricart:
[FEAT] added an expirations command, which traverses the current operator checking for JWTs and creds that expire within a specified amount of time (by default, a month). The command can generate a report in JSON format that could be used by other tooling.


derekcollison (approved):
LGTM

PR #624, opened 2023-11-13
[FIX] generate --mem-resolver performance in setups that have huge number of users

aricart:
[FIX] generate command has a generic processing engine but when generating anything but nkey configurations, it shouldn't look at the users as this could significantly degrate generation performance when that data is not needed.


ripienaar (approved):
LGTM

PR #634, opened 2024-02-01
Enhancements to edit signing-key

### LINKED ISSUE #633: Add an option to create a scoped signing-key in one command

Currently, to create a scoped signing-key, one hase to call `nsc --sk generate`, note the (non machine parseable) ouput, then call the command `nsc --sk KEYJUSTCREATED --role blah`.

Allow the use of the `--role` parameter when using `nsc --sk generate`, so the key is immediately scoped, and can be referred by its role further in the key deployment process.

Using decentralized authentication for NATS is a very good feature. Scoped signing-keys is even better.
The current way the creation is implemented is error prone, because you either depend on one's copy/paste skills, or parsing an stdout output which you have no warranty that it won't change in the future.

I need to check the generate return code, and if it is valid, grep / awk / sed the output to get the key argument to pass to the next nsc call. Very inefficient.

aricart:
[FEAT] changed `edit signing-key` to allow `--sk generate`, or infer the sk if the value for the role is given or just the role is given

FIX #633

### DIFF
--- a/cmd/editscopedsk.go
+++ b/cmd/editscopedsk.go
@@ -40,8 +40,8 @@ nsc edit signing-key --account <n> --sk <sk> --allow-sub <subject>,...
 			return RunAction(cmd, args, &params)
 		},
 	}
-	cmd.Flags().StringVarP(&params.skName, "sk", "", "", "signing key to set scope for or role name for already existing scoped signing key")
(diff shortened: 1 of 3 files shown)


ripienaar (approved):
LGTM

PR #638, opened 2024-02-27
[FEAT] edit w3c tracing context configurations on account and enable tracing into service exports

### LINKED ISSUE #636: add compatibility with tracing

https://github.com/nats-io/jwt/pull/216

above

_No response_

_No response_

aricart:
Fix #636

### DIFF
--- a/cmd/editaccount.go
+++ b/cmd/editaccount.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2023 The NATS Authors
+ * Copyright 2018-2024 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
@@ -121,6 +121,10 @@ func createEditAccount() *cobra.Command {
 
 	params.signingKeys.BindFlags("sk", "", nkeys.PrefixByteAccount, cmd)
 	params.TimeParams.BindFlags(cmd)
+
+	cmd.Flags().StringVarP(&params.traceContextSubject, "trace-context-subject", "", "trace.messages", "sets the subject where w3c trace context information is sent. Set to \"\" to disable")
+	cmd.Flags().VarP(&params.traceContextSampling, "trace-context-sampling", "", "set the trace context sampling rate (1-100) - 0 default is 100")
+
 	return cmd
 }
 
@@ -148,22 +152,24 @@ type EditAccountParams struct {
 	GenericClaimsParams
 	PermissionsParams
 	JetStreamLimitParams
-	claim            *jwt.AccountClaims
-	token            string
-	infoUrl          string
-	description      string
-	conns            NumberParams
-	leafConns        NumberParams
-	exports          NumberParams
-	disallowBearer   bool
-	exportsWc        bool
-	imports          NumberParams
-	subscriptions    NumberParams
-	payload          NumberParams
-	data             NumberParams
-	signingKeys      SigningKeysParams
-	rmSigningKeys    []string
-	disableJetStream bool
+	claim                *jwt.AccountClaims
+	token                string
+	infoUrl              string
+	description          string
+	conns                NumberParams
+	leafConns            NumberParams
+	exports              NumberParams
+	disallowBearer       bool
+	exportsWc            bool
+	imports              NumberParams
+	subscriptions        NumberParams
+	payload              NumberParams
+	data                 NumberParams
+	signingKeys          SigningKeysParams
+	rmSigningKeys        []string
+	disableJetStream     bool
+	traceContextSubject  string
+	traceContextSampling NumberParams
 }
 
 func (p *EditAccountParams) SetDefaults(ctx ActionCtx) error {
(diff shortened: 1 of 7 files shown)


cmd/editaccount.go
```diff
@@ -818,6 +876,28 @@ func (p *EditAccountParams) Run(ctx ActionCtx) (store.Status, error) {
...
+	// if they provided the flag we change it
+	if flags.Changed("trace-context-subject") {
+		// if they set us to "", we disable it
+		if p.traceContextSubject == "" {
+			p.claim.Trace = nil
+			r.AddOK("disabled trace context")
+		} else {
+			p.claim.Trace = &jwt.MsgTrace{}
+			p.claim.Trace.Destination = jwt.Subject(p.traceContextSubject)
+			r.AddOK("changed trace context subject to %d", p.claim.Trace.Sampling)
```
kozlovic, line 888: You are supposed to show the change of the subject, but you use `%d` and show the sampling value.

cmd/editaccount.go
```diff
@@ -818,6 +876,28 @@ func (p *EditAccountParams) Run(ctx ActionCtx) (store.Status, error) {
...
+		}
+	}
+
+	if flags.Changed("trace-context-sampling") {
+		if p.claim.Trace == nil {
+			return r, errors.New("cannot set context sampling rate when disabling the trace context")
+		}
+		// we already validated that the subject is there either existing or new
+		p.claim.Trace.Sampling = int(p.traceContextSampling.Int64())
+		r.AddOK("changed trace context sampling to %d", p.claim.Trace.Sampling)
```
kozlovic, line 898: You could add `to %d%%` to show it as a percentage, but it is ok this way too.

kozlovic (changes requested):
Some changes needed, but also you need to add the allow trace for the stream import (with similar checks that will fail allow-trace on service import). See JWT pr for details: https://github.com/nats-io/jwt/pull/216

cmd/addimport.go
```diff
@@ -542,7 +552,10 @@ func (p *AddImportParams) createImport() *jwt.Import {
 	if p.service {
 		im.Type = jwt.Service
 		im.Share = p.share
+	} else if p.allowTrace {
+		im.AllowTrace = true
```
kozlovic, line 556: Are you sure it would not be `im.AllowTrace = p.allowTrace` instead of `true`?

cmd/addimport.go
```diff
@@ -444,6 +447,13 @@ func (p *AddImportParams) PostInteractive(ctx ActionCtx) error {
 		return err
 	}
 
+	if !p.service {
+		p.allowTrace, err = cli.Confirm("allow tracing", false)
```
kozlovic, line 451: For service export, you did that in "PreInteractive" and here you are doing for "PostInteractive", any reason for that?

cmd/addimport.go
```diff
@@ -542,7 +552,10 @@ func (p *AddImportParams) createImport() *jwt.Import {
 	if p.service {
 		im.Type = jwt.Service
 		im.Share = p.share
+	} else if p.allowTrace {
+		im.AllowTrace = true
 	}
+
```
kozlovic, line 558: I don't see also where you set the defaults like you did for the service export.

cmd/addimport_test.go
```diff
@@ -554,4 +558,5 @@ func TestAddImport_SameName(t *testing.T) {
 	require.Equal(t, "stream.database", string(ac.Imports[1].Subject))
 	require.True(t, ac.Imports[1].IsStream())
 	require.Equal(t, ts.GetAccountPublicKey(t, "B"), ac.Imports[1].Account)
+	require.True(t, ac.Imports[1].AllowTrace)
 }
```
kozlovic, line 562: Would it make sense to add the tests similar to what you did for service export here for stream import. I am referring to TestAddServiceExportWTracing and TestAddStreamExportWTracing.

kozlovic (changes requested):
Some issues and tests missing. Also, do we need to add an "edit import" file? If not, does it mean that currently you can't modify a stream import, you would need to do what? delete and recreate?

cmd/addimport_test.go
aricart: No because those prompts won't show unless the import/export is the right kind.

cmd/addimport.go
aricart: works differently on the other you can edit, so it has to have some default value. On the import it is one shot, you cannot edit.

cmd/addimport.go
aricart: Yes, in the case where they are creating the import from an activation token or by picking from some export, some of the options may or not exist. By PostInteractive() the question can be asked for that.

cmd/addimport.go
aricart: adding an import is more complex than an export, because they can simply provide an activation.

cmd/addimport.go
kozlovic: Oh sorry... I saw the `else` of the `if p.service` and I thought you were just processing stream properties here and always setting to `true` without seeing that the `else if p.allowTrace`. My bad.

cmd/addimport.go
kozlovic: ok

cmd/addimport.go
kozlovic: ok

cmd/addimport_test.go
kozlovic: ok

kozlovic (approved):
LGTM

PR #645, opened 2024-04-18
[FEAT] added `--description` to scoped signing keys for storing arbitrary scope information

aricart:
[FEAT] describe account now lists the scope description


ripienaar (approved):
LGTM

PR #656, opened 2024-05-30
Fixes related to --require-signing-key option in operator

### LINKED ISSUE #628: Bad diagnostic for adding user without using signing-key when operator is strict_signing_key_usage

nsc 2.8.4.

Linux/amd64 (Ubuntu 20.04 LTS)

Yes.

1. Have an operator with the `strict_signing_key_usage` set to true.  Optionally, forget that this is the case.
2. Create a new account.
3. Try to create a user in the account.
4. See an error about "any of the following" with nothing following

A diagnostic which points me in the right direction, along the lines of:

```
error: operator 'foobar' has strict_signing_key_usage set, refusing to add user to account 'baz' without an account signing key
```

```console
$ nsc add user --name pt-second-pdp-2023-12-06
Error: unable to resolve any of the following signing keys in the keystore:
[error code 1] $
```

aricart:
[FIX] added a guard to edit operator --require-signing-key to ensure that operator has at least one signing key before setting it

[FIX] added a guard when removing signing keys from the operator that require signing key is not set [FIX] added better messages in cases where a signing key is required because of --require-signing-key option when adding users

Fixes #628


scottf (approved):
LGTM

PR #658, opened 2024-07-05
[FIX] mapping add/delete commands related to the `--cluster` flag

### LINKED ISSUE #657: NSC does not allow multiple clusters with 100% weight subject mapping

nsc version 2.8.6

darwin-arm64

Run the following:
```
nsc add account APP
nsc edit account APP --sk generate
nsc add mapping --account APP --cluster east --from "foo" --to "foo.east" --weight 100
nsc add mapping --account APP --cluster west --from "foo" --to "foo.west" --weight 100
```

That the mapping command would add another element to the mapping slice instead of not allowing it. Trying to do the following with nsc: https://github.com/synadia-labs/eventually-consistent-virtual-global-stream/blob/main/mappings.cfg
But so far no luck

When the commands run I get the following:
```
$ nsc add mapping --account APP --cluster east --from "foo" --to "foo.east" --weight 100
[ OK ] added mapping foo -> foo.east, weight 100%
$ nsc add mapping --account APP --cluster west --from "foo" --to "foo.west" --weight 100
Error: mapping validation failed: [Mapping "foo" exceeds 100% among all of it's weighted to mappings]
```

aricart:
- [BUMP] bumped dependencies (including JWT to support cluster allocation on checks)
- [FIX] fixed an issue in add mapping where the `--cluster` was not taken into account when deciding whether to add or replace, and the first mapping matching to `--to` was replaced. 
- [FEAT] delete mapping now supports --cluster

Fixes #657 - Note that this fix [requires the server to adopt the updated JWT](https://github.com/nats-io/nats-server/pull/5618) library as otherwise, validation by the server will reject the mapping configuration


jnmoyne (approved):
LGTM

PR #665, opened 2024-09-11
[FIX] edit signing-key was unable to remove a connection type when in lowercase

maxarndt:
This change ensures that connection types are uppercased, and autofixes any type that is lowercase in an existing scoped signing-key.

In reference to #664 

Please review carefully, this is my first contribution to `nsc`.

### DIFF
--- a/cmd/editscopedsk_test.go
+++ b/cmd/editscopedsk_test.go
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/nats-io/jwt/v2"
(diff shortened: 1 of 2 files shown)


cmd/editscopedsk_test.go
```diff
@@ -221,3 +222,123 @@ func Test_EditScopedSkByRole(t *testing.T) {
...
+	_, _, err = ExecuteCmd(createEditSkopedSkCmd(), "--sk", "foo", "--conn-type", "bar")
+	require.Error(t, err)
+
+	// add lower case conn type - this is prevented now, but worked in the past
+	ac, err := ts.Store.ReadAccountClaim("A")
+	require.NoError(t, err)
+	scope, ok := ac.SigningKeys.GetScope(ac.SigningKeys.Keys()[0])
+	require.True(t, ok)
+	scope.(*jwt.UserScope).Template.AllowedConnectionTypes.Add(strings.ToLower(jwt.ConnectionTypeStandard))
+	ac.SigningKeys.AddScopedSigner(scope)
```
maxarndt, line 249: Was not able to see any impact of calling `AddScopedSigner()` here. But it is used [here](https://github.com/nats-io/nsc/blob/92dd1350897317c7903f79d6127ed146a14f1403/cmd/editscopedsk.go#L194) so I wanted to double check if this should be kept or removed from this test.

aricart (approved):

cmd/editscopedsk_test.go
aricart: I think the test is fine.

maxarndt:
@aricart thanks for the review! I rebased my branch. You can kick-off the pipelines again.

PR #668, opened 2024-09-12
Tags are now case-sensitive, describe account lists tags associated with the account

aricart:
[CHANGE] Tags (--tag and --rm-tag) are now case-sensitive
[FIX] describe accounts was not printing tags associated with the account

Fixes https://github.com/nats-io/nats-server/issues/5681

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -52,3 +52,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/nats-io/jwt/v2 => github.com/nats-io/jwt/v2 v2.6.1-0.20240912200904-90d24c5a61af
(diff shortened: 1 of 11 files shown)


go.mod
```diff
@@ -52,3 +52,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/nats-io/jwt/v2 => github.com/nats-io/jwt/v2 v2.6.1-0.20240912200904-90d24c5a61af
```
ripienaar, line 56: probably can't merge this?

ripienaar (approved):
Looks ok, what about the potential breaking impact?

aricart:
> Looks ok, what about the potential breaking impact?

by itself nothing - the tags will be all lowercase on anything loaded. However, sloppy input on the cli will possibly create unexpected results.

aricart:
After trying a couple of options settled on the change that JWT API uses are always case-sensitive as that can be telegraphed through the release notes. But for `nsc`, added the flag `--strict-tags`, which allows non-lowercased values for tag operations. 

This will catch sloppy usages of the flag where values are not lowercase and request that the `--strict-tags` option be specified. The other check is that the JWT will now report errors when trying to remove tags that are not present.

ripienaar (approved):
LGTM

PR #679, opened 2024-12-12
Enhance operator reissue command with key support

aricart:
Add support for using a private key during operator reissue, enabling identity updates directly from operator seeds. Updated error handling for key loading and added corresponding tests to validate functionality and ensure accounts are correctly re-signed.


wallyqs (approved):
LGTM

PR #683, opened 2024-12-16
Generate docs on every commit

alexbozhenko:
https://github.com/nats-io/nsc/tree/main/docs
has not been updated for a while:

Can we generate it automatically in every PR, 
similarly to 
https://github.com/ConnectEverything/nats-by-example/blob/e3db9b0e98f2010390ba3dfb02c108ac8accf713/.github/workflows/deploy.yaml#L50C1-L60C71


alexbozhenko:
TODO drop date from generation

aricart (approved):
LGTM

PR #701, opened 2025-01-15
Add dependabot

alexbozhenko:
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference

Same as in https://github.com/nats-io/nats-server/blob/74f005dd2887c441ea3d9680c8dcbaa500484ee9/.github/dependabot.yml#L4


alexbozhenko:
cc @aricart

aricart (approved):
LGTM

PR #704, opened 2025-01-21
import/export of operator

aricart:
command to export an operator and all related subentities and their keys as JSON.
Also import that reads the file and creates the operator optionally renaming it and changing its keys


scottf (approved):
LGTM

BZValoche:
Why is this still not merged?

PR #724, opened 2025-03-28
cmd: add nil check for pull report

wooffie:
In logic of checks inside loop we should add check, because if j.PullStatus == nil and j.err == nil - we do nil dereference


aricart:
@wooffie are you getting a nil dereference? The status code doesn't have anything to do with that, it could be an error/warning/etc and the filter would hide the report.

aricart:
@wooffie  If you are getting a nil dereference please include the stack trace.

wooffie:
No, I don't get panic, just triaging sast result, and thought that we can get it here.

aricart:
You cannot.

PR #725, opened 2025-03-28
Add stricter validation for signing keys in push/pull commands

### LINKED ISSUE #719: during push nsc may try to create dynamic sys user against scoped signing key, which will fail

Latest 2.10.2

Linux

Yes

When selecting a signing key for creating a dynamic sys user during push. nsc should check:

1. Not a scoped signing key

The customer (anonymous) has created a scoped signing key and a SYS user for use in leaf node SYS account connections. This SYS user establish unidirectional trust. For extensibility and standardization a scoped signing key is used for that leaf node system user.

When do an nsc push, the method `push.go -> getSystemAccountUser()` will may pick the first signing key, without checking if it is scoped. When trying to create a temporary sys user against the scoped signing key it may:

1. Fail with: "User JWT is not valid: scoped users require no permissions or limits set"
2. OR even if it succeeds the user may not have enough permission to execute the push.

Workaround: Create and user an explicit unrestricted System user.

aricart:
[CHANGE] for accounts that have operators with strict signing keys, if the account only uses scoped signing key - the command will require the system user to be specified, previously it selected the first key, which may or not have the correct permissions.

This update adds comprehensive tests and improved validation logic to handle signing keys when pushing or pulling configurations. The changes ensure proper differentiation between scoped and non-scoped keys and provide clearer error messages when required keys are missing or improperly configured.

Fixes #719

### DIFF
--- a/cmd/push.go
+++ b/cmd/push.go
@@ -124,6 +124,7 @@ func getSystemAccountUser(ctx ActionCtx, sysAccName, sysAccUserName, allowSub st
 			return "", nil, fmt.Errorf(`system account "%s" not found`, op.SystemAccount)
 		}
 	}
+
 	getOpt := func(theJWT string, kp nkeys.KeyPair) nats.Option {
 		return nats.UserJWT(
 			func() (string, error) {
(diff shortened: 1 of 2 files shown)


scottf (approved):
LGTM

PR #798, opened 2026-03-09
Add short aliases for list subcommands

joeriddles:
Allow `nsc list operator`, `nsc list account`, `nsc list user`, and `nsc list key` as aliases for their plural counterparts.


aricart:
maybe have duplicate one of the tests where it uses the aliases ty.

aricart (approved):
LGTM

PR #821, opened 2026-05-20
feat: add FIPS 140-3 compliance support (#812)

### LINKED ISSUE #820: Allow for a FIPS-140 only build mode

Go 1.24 added a FIPS 140-3 validated cryptographic module, activated at runtime with `GODEBUG=fips140=only`. In that mode, calls to non-approved algorithms return errors. We want an `nsc` build that runs cleanly under `fips140=only` so users with compliance requirements can use it.

The blocker is that nkeys curve keys (xkeys) are X25519, which is not in Go's FIPS module. Under `fips140=only`, any path that generates a curve keypair will fail at the crypto layer. Identity nkeys (Ed25519) and JWT signing are fine, both are in the validated boundary.

Separately, `nsc update` uses `go-github-selfupdate`, whose signature verification path is not audited for FIPS. A FIPS-compliant deployment shouldn't pull unverified binaries anyway.

1. New package `internal/fips` with `Enabled()` and `Enforced()` wrappers around `crypto/fips140`, plus a `HardenTLS(*tls.Config)` helper that sets `MinVersion: TLS12` and restricts `CurvePreferences` to P-256/P-384/P-521.

2. Refuse curve keypair generation in FIPS mode at the two call sites:
   - `cmd/signingkeyseditor.go`, function `resolve` (line 63), before `nkeys.CreatePair(e.kind)` at line 68. Catches `nsc edit authorization --curve generate` and any other caller using `SigningKeysParams` with `PrefixByteCurve`.
   - `cmd/generatenkey.go`, the `PrefixByteCurve` switch arm at line 74, before `nkeys.CreatePair(e.prefix)` at line 82.

   Hide `--curve` / `--curves` from help output when FIPS is enforced.

3. Short-circuit `nsc update` in `cmd/update.go` before the call to `selfupdate.DefaultUpdater().UpdateTo` at line 94 with a message pointing at manual install.

4. In `cmd/tool.go`, function `createDefaultToolOptions` (line 48), append `nats.Secure(fips.HardenTLS(nil))` when FIPS is enabled. Gives a clearer error than a generic TLS failure when connecting to a non-FIPS server.

## Verification

`GOFIPS140=v1.0.0 GODEBUG=fips140=only go test ./...` should pass with curve and xkey tests skipped via build tag or runtime skip.

_No response_

_No response_

aricart:
- Introduced FIPS enforcement checks across tools and commands.
- Added `skipIfFIPS` utility to disable non-compliant tests and operations.
- Updated TLS configuration to allow NIST curve preferences in strict FIPS mode.
- Disallowed WebSockets under strict FIPS until `nats.go` fixes SHA-1 dependency.
- Bumped dependencies (`nats-server`, `nats.go`, `compress`, etc.).

Fixes #820


aricart:
@claude review this

mtmk (approved):
LGTM

